PR #1, opened 2016-03-21
java.verify_ver does not verify java-1.8.0 (SOFTWARE-2212)

brianhlin:
The output of `java -version` differs between java-1.7.0 and java-1.8.0. The former has its version number preceded by "java version" and the latter has its version preceded by "openjdk version". This commit changes `java.verify_ver()` to look for the version in the first line, ignoring any preceding text.

OpenJDK 1.7.0
```
# java -version
java version "1.7.0_95"
OpenJDK Runtime Environment (rhel-2.6.4.0.el7_2-x86_64 u95-b00)
OpenJDK 64-Bit Server VM (build 24.95-b01, mixed mode)
```

OpenJDK 1.8.0
```
# java -version
openjdk version "1.8.0_71"
OpenJDK Runtime Environment (build 1.8.0_71-b15)
OpenJDK 64-Bit Server VM (build 25.71-b15, mixed mode)
```

### DIFF
--- a/osgtest/library/java.py
+++ b/osgtest/library/java.py
@@ -34,14 +34,14 @@ def select_ver(java_type, version):
 def get_ver(java_type):
     """Find version of java that is currently selected in the alternatives"""
     current_java = _run_alternatives(java_type, '\n', 'find java version for selection')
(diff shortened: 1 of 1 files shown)


osg-cat:
Maybe it would be a good idea still to look for the word “version” (then whitespace) prior to a version string?

PR #6, opened 2016-04-21
The tests of gratia meter should not depend on gram and globus packag…

efajardo:
The tests of gratia should not depend on some globus and gram packaing installed.

### DIFF
--- a/osgtest/tests/test_54_gratia.py
+++ b/osgtest/tests/test_54_gratia.py
@@ -411,8 +411,7 @@ class TestGratia(osgunittest.OSGTestCase):
     #This test executes condor_meter
     def test_18_execute_condor_meter(self):
         core.state['gratia.condor-meter-running'] = False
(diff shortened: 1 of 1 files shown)


efajardo:
assigned to @brianhlin

osg-cat:
I am not convinced that this change is adequate… only a few lines down, the code checks to make sure that a Globus gatekeeper is running! Will someone please review this module carefully, understand what is and is not required for it to succeed, and update accordingly?

brianhlin:
Have you tested the changes? 

> If you need to change osg-test test modules, patches go into test-changes.patch. These patches can be generated with git diff. 

https://twiki.opensciencegrid.org/bin/view/SoftwareTeam/TestRunsAsVMs

efajardo:
Hi @brianhlin ,

I agree this has to be more carefully thought. Right now the way it is only checks for GRAM. It should do a check for a Gram or HTCOndorCe installed. SO I am closing this pull request and go back to the drwaing board for some tests.

PR #7, opened 2016-04-25
SOFTARE-2293 execute condor_meter without gram

efajardo:
@brianhlin can your please review this? I manually tested on a vm and it works fine.

### DIFF
--- a/osgtest/tests/test_54_gratia.py
+++ b/osgtest/tests/test_54_gratia.py
@@ -411,7 +411,8 @@ class TestGratia(osgunittest.OSGTestCase):
     #This test executes condor_meter
     def test_18_execute_condor_meter(self):
         core.state['gratia.condor-meter-running'] = False
-        core.skip_ok_unless_installed('gratia-probe-condor', 'gratia-service', 'osg-ce')
+        core.skip_ok_unless_installed('gratia-probe-condor', 'gratia-service')
+        core.skip_ok_unless_one_installed('htcondor-ce-condor', 'globus-gram-job-manager-condor')
         self.skip_bad_if(core.state['gratia.condor-logs-copied'] == False)
         self.skip_bad_unless(core.state['globus-gatekeeper.running'] or core.state['condor-ce.started'], 'gatekeeper not running')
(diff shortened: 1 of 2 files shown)


osgtest/tests/test_54_gratia.py
```diff
@@ -411,10 +411,9 @@ def test_17_copy_condor_logs(self):
     #This test executes condor_meter
     def test_18_execute_condor_meter(self):
         core.state['gratia.condor-meter-running'] = False
-        core.skip_ok_unless_installed('gratia-probe-condor', 'gratia-service', 'globus-gram-job-manager-condor',
-                                      'globus-gram-client-tools', 'globus-proxy-utils')
+        core.skip_ok_unless_installed('gratia-probe-condor', 'gratia-service', 'osg-ce')
```
brianhlin: With [SOFTWARE-2278](https://jira.opensciencegrid.org/browse/SOFTWARE-2278), doesn't this mean that this test will not run in the GRAM gatekeeper case? You should check out `core.skip_ok_unless_one_installed()` and list both htcondor-ce and the globus-gatekeeper packages.

efajardo: That was my original thought. But from talking with @osg-cat, I thought the metapackage osg-ce is used to install either of them. So it would be enough to check for it.

efajardo:
@brianhlin I made the change you requested. Can you please review this?

brianhlin:
Looks solid to me. Could you kick off a VMU test run and throw the diff into `test-changes.patch`? We should see the `osg-testing` results go green if this works.

bbockelm:
Just a thought: there's no reason why we can't run a subset of the VMU tests in travis-ci.  Should help with the iteration speed.

efajardo:
Finally the tests are green 👍 

http://vdt.cs.wisc.edu/tests/20160427-1517/results.html
@brianhlin can you merge it now?

brianhlin:
The `osg-ce-pbs` cleanup errors came about after the changes to `osg-ce` (http://vdt.cs.wisc.edu/tests/20160420-1554/results.html) and not the `osg-test` changes so this looks good.

PR #8, opened 2016-04-28
osg-test should exit != 0 if tests fail (SOFTWARE-2306)

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-2306

Tests look good other than errors that are being fixed in #7: http://vdt.cs.wisc.edu/tests/20160427-1413/results.html


bbockelm:
Looks good to me.

PR #10, opened 2016-04-28
Add option to osg-test that turns on SELinux (SOFTWARE-2270)

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-2270

Pass `--selinux` to set SELinux to `enforcing` from `permissive` via `setenforce`. I spot checked the tests to verify that they're correctly enabling SELinux (or not in the first case): 

* VMU tests with SELinux disabled (default): http://vdt.cs.wisc.edu/tests/20160428-1329/results.html
   * 3.3 ALL tests are due to #7, which was merged after these tests were kicked off
   * Random errors appear unrelated
* VMU tests with SELinux enabled: http://vdt.cs.wisc.edu/tests/20160428-1335/results.html
   * 3.3 ALL tests are due to #7, which was merged after these tests were kicked off
   * EL6 CMVFS and cleanup test failures, presumably due to enabled SELinux
   * EL7 HTCondor-CE, RSV, and Globus-Condor job submission failures, presumably due to enabled SELinux

### DIFF
--- a/osg-test
+++ b/osg-test
@@ -30,6 +30,7 @@ def parse_command_line():
                 "packages": [],
                 "password": 'vdttest',
                 "printtest": True,
+                "selinux": False,
                 "securepass": False,
                 "skip_cleanup": False,
                 "skiptests": False,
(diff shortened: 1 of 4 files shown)


brianhlin:
@osg-cat: assigning to you for now since I'm ooo tomorrow. Feel free to delegate.

osg-cat:
In the future, please separate pull requests with formatting-only commits from functional ones.

In this pull request, in special_cleanup, should you renumber the methods following the inserted one?

Also, should the checks for installed RPMs in special_cleanup and special_selinux be the same?

brianhlin:
* Fixed numbering in the cleanup tests
* Checks for installed RPMs are different because we use two SELinux utilities in special_selinux (`sestatus` and `setenforce`) but only need to use one (`sestatus`) in special_cleanup.

PR #11, opened 2016-05-02
Run gfal2util tests before gums/condor

brianhlin:
Test results look good here: http://vdt.cs.wisc.edu/tests/20160502-1014/results.html

### DIFF
--- a/files/test_sequence
+++ b/files/test_sequence
@@ -35,12 +35,12 @@ test_49_jobs
 test_50_voms
 test_51_edgmkgridmap
 test_52_bestman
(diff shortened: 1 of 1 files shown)


osg-cat:
We just need to remember that we made this change to avoid (for now) GridFTP authentication issues caused by tests between test_52_bestman and test_58_gfal2utils (when in their logical sequence).

brianhlin:
I've created [SOFTWARE-2320](https://jira.opensciencegrid.org/browse/SOFTWARE-2320) to investigate the cause of the failing gfal tests.

PR #12, opened 2016-05-02
Fix bug that caused missing test logs for successful runs

brianhlin:
Test results here: http://vdt.cs.wisc.edu/tests/20160502-1029/results.html

Logs are back!


osg-cat:
I am still not crazy about a sub-function making a decision about the user interface (in this case, an exit code). But this code should work for now, so I will accept it.

PR #14, opened 2016-05-03
Move exit code logic to osg-test main

brianhlin:
This is not intended for the May release but I figured I'd fix it before we forgot about it. Here's the good test output (other than random failures and ignored fetch-crl failures) using my branch: http://vdt.cs.wisc.edu/tests/20160502-1554/results.html


osg-cat:
Yeah, I think that is much clearer.

PR #17, opened 2016-05-12
Add osg-ca-generator check in case osg-test has been installed from s…

brianhlin:
…ource


edquist:
Looks good to me.

I did not know you can do sys.exit(error_message) before - neat!

PR #18, opened 2016-05-12
add timeout option to core.system() (SOFTWARE-646)

edquist:
In core, add a per-command timeout option to __run_command(), system(),
and check_system().

This will remain unused by default; you have to explicitly pass
timeout=NN[.NN] to core.system() or core.check_system() in your tests if
you want to use it.

a TERM signal is used by default to kill the process group of
subprocesses that exceed their timeout.  Use timeout_signal='KILL', etc,
if some other signal is desired.


brianhlin:
After learning more about fork/exec/waitpid, I'm mostly satisfied with the changes but I do have questions:

* If I'm reading this correctly, `check_system` will fail on timeouts but will not differentiate between timeout and other failures. I think we should probably edit the failure message in the case of timeouts (and maybe create a specific timeout type of test failure status i.e. badskip?)
* Have you run a timeout test in osg-test to make sure the TestRunner handles everything properly?

edquist:
Ah!  You're right; while the status returned by `core.system` will be `-1` on timeout, `check_system` doesn't do anything specific with it -- the failure message will include `EXIT STATUS: -1`, but (I'm guessing) it would be clearer if the failure message specifically said it timed out.  (We can add this in `core.diagnose` by checking if `status == -1`.)

(So, to repeat: yes, currently the only way for the caller of `check_system` to detect a timeout would be to see if the failure message matches `EXIT STATUS: -1`.)

But, currently the only way `check_system` communicates a failed check to the caller is by the failed assert throwing an AssertionError, right?  So as for the caller detecting that `check_system` had a timeout, I suppose we could have `check_system` throw a new exception type (TimeoutError or something).  I'm open to suggestions.

I haven't tested adding timeouts to any tests yet, no :wink:  ...I can get back to you on that...

brianhlin:
We could create a new `TimeoutError` exception based on `AssertionError` (see [osgunittest.py](https://github.com/opensciencegrid/osg-test/blob/master/osgtest/library/osgunittest.py#L20-L34)) so that it plays nicely with the unittest framework. This would also mean updates to the test framework to accommodate the new test result state in the reporting. Alternatively, we can just treat them as failures and check if `status == -1` and prepend the user's message with `TIMEOUT` or some such similar text.

Opinions @osg-cat, @matyasselmeci ?

osg-cat:
I <3 clarity, which I think a separate exception class achieves. We can always choose to treat that new exception in the same way as others.

edquist:
I now have the timeout watcher process call `_exit` instead of regular `exit` or get killed with `SIGKILL` to avoid running any unittest exit handlers - @brianhlin, hopefully this will fix the double-output issue you were seeing...

brianhlin:
Test results look good (http://vdt.cs.wisc.edu/tests/20160601-1442/results.html). Manual tests (adding a command that timed out) failed properly.

PR #22, opened 2016-05-25
Limit the number of yum cleans that we perform in the tests (SOFTWARE…

brianhlin:
…-2337)

Successful test results [here](http://vdt.cs.wisc.edu/tests/20160525-1011/results.html). Basically yum cleans are only performed on yum failures instead of at the beginning of each yum retried command.

### DIFF
--- a/osgtest/library/yum.py
+++ b/osgtest/library/yum.py
@@ -10,11 +10,6 @@ import time
 
 import osgtest.library.core as core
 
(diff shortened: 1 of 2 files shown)


brianhlin:
We're seeing lots of RHEL install failures again because of subscription manager stuff and I think these changes could help so I'm merging this.

PR #24, opened 2016-06-03
Fix detection of Tomcat startup in osg-test

brianhlin:
http://vdt.cs.wisc.edu/tests/20160602-1452/results.html

Test results (look good): http://vdt.cs.wisc.edu/tests/20160602-1452/results.html

### DIFF
--- a/osgtest/tests/test_24_tomcat.py
+++ b/osgtest/tests/test_24_tomcat.py
@@ -42,10 +42,9 @@ class TestStartTomcat(osgunittest.OSGTestCase):
     def test_05_start_tomcat(self):
         core.skip_ok_unless_installed(tomcat.pkgname())
         core.state['tomcat.started'] = False
-        core.config['tomcat.catalina-out'] = '/var/log/%s/catalina.out' % tomcat.pkgname()
 
         try:
-            initial_stat = os.stat(core.config['tomcat.catalina-out'])
+            initial_stat = os.stat(tomcat.catalinafile())
         except OSError:
             initial_stat = None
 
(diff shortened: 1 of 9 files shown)


osgtest/tests/test_76_tomcat.py
```diff
@@ -28,6 +27,11 @@ def test_03_deconfig_tomcat_properties(self):
         core.skip_ok_unless_installed(tomcat.pkgname(), 'emi-trustmanager-tomcat')
         files.restore(os.path.join(tomcat.sysconfdir(), 'server.xml'), 'tomcat')
 
+    def test_04_deconfig_catalina_logging(self):
+        core.skip_ok_unless_installed(tomcat.pkgname())
+        if core.el_release() == 5: # limit okskip list
```
osg-cat: I don’t understand this comment in this context…

osgtest/tests/test_24_tomcat.py
```diff
@@ -43,12 +34,39 @@ def test_04_config_tomcat_endorsed_jars(self):
             new_contents = old_contents + "\n" + line
             files.write(tomcat.conffile(), new_contents, owner='tomcat')
 
+    def test_04_configure_gratia(self):
+        core.skip_ok_unless_installed(tomcat.pkgname(), 'gratia-service')
+        command = ('/usr/share/gratia/configure_tomcat',)
+        core.check_system(command, 'Unable to configure Gratia.')
+
     def test_05_start_tomcat(self):
```
osg-cat: Is it absolutely critical to have a Gratia step mingled in with the generic Tomcat steps?

osg-cat:
Overall, I think that this set of changes looks good, but I had one substantive comment that I would like addressed before accepting the whole mess.

osgtest/tests/test_24_tomcat.py
brianhlin: Yes: including it in `test_23_gratia.py` results in config changes being overwritten by the earlier tests in `test_24_tomcat.py`.

brianhlin:
I've updated the comment with hopefully a clearer one: https://github.com/opensciencegrid/osg-test/pull/24/commits/a1eeec146034663c93dc8d6e9764c9b9e770476f

I think my comment about the gratia config test in `test_24_tomcat.py` should address your concerns.

osg-cat:
OK, the disentangling of Tomcat and Gratia steps will have to wait until we split apart the test modules more.

PR #26, opened 2016-06-08
Fix Gratia server.xml.template on EL7

brianhlin:
#### Problem

GUMS + ALL EL7 tests were [broken](http://vdt.cs.wisc.edu/tests/20160603-1216/results.html) due to the incorrect `sSLImplementation` variable in the `server.xml.template` that Gratia uses in its tomcat configuration script: it needed to be `sslImplementationName`. Curiously, GUMS on EL5 and EL6 [broke](http://vdt.cs.wisc.edu/tests/20160606-1526/results.html) unless I used `sSLImplementation`. Sigh.

#### Solution

The successful run is [here](http://vdt.cs.wisc.edu/tests/20160607-1646/results.html) after fixing the above and a strange requirement of the `replace_regexpr()` function where the new line had to match the regex being replaced.

Do we also need to update osg-tested-internal on EL7 to now include GUMS?

### DIFF
--- a/osgtest/library/files.py
+++ b/osgtest/library/files.py
@@ -144,12 +144,9 @@ def replace(path, old_line, new_line, owner=None, backup=True):
     write(path, lines_to_write, owner, backup)
 
 def replace_regexpr(path, regexp, new_line, owner=None, backup=True):
-    """Replace an old line that matches regexpr with a new line in given path.                                                     
-    The 'owner' and 'backup' arguments are passed to write().                                                                                  
+    """Replace an old line that matches regexpr with a new line in given path.
+    The 'owner' and 'backup' arguments are passed to write().
     """
-    match = re.search(regexp, new_line)
-    if not match:
-       raise ValueError("The new_line '%s' does not match reg expr '%s'" % (new_line, regexp)) 
     lines_to_write = []
     lines = read(path)
     for line in lines:
--- a/osgtest/tests/test_23_gratia.py
+++ b/osgtest/tests/test_23_gratia.py
@@ -1,8 +1,10 @@
 import os
 import shutil
+
 import osgtest.library.core as core
 import osgtest.library.files as files
 import osgtest.library.osgunittest as osgunittest
+import osgtest.library.tomcat as tomcat
 
 class TestStartGratia(osgunittest.OSGTestCase):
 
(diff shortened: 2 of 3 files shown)


osgtest/library/files.py
```diff
@@ -144,12 +144,9 @@ def replace(path, old_line, new_line, owner=None, backup=True):
...
 def replace_regexpr(path, regexp, new_line, owner=None, backup=True):
-    """Replace an old line that matches regexpr with a new line in given path.                                                     
-    The 'owner' and 'backup' arguments are passed to write().                                                                                  
+    """Replace an old line that matches regexpr with a new line in given path.
+    The 'owner' and 'backup' arguments are passed to write().
     """
-    match = re.search(regexp, new_line)
-    if not match:
-       raise ValueError("The new_line '%s' does not match reg expr '%s'" % (new_line, regexp)) 
     lines_to_write = []
```
osg-cat: Hm, I am not sure that I agree with this change. At very least, I would have made the match optional and, via an argument, made it required by default. I think that the point was that one is *expecting* the match to work and hence the change to take place.

Another possibility is to have the function return the number of changed lines, such that the caller could examine the return and decide if 0 (no lines changed) is acceptable — but of course both the PR change and this idea change the semantics of this function and thus require going through test code to see if that is OK.

osgtest/tests/test_23_gratia.py
```diff
@@ -164,3 +166,12 @@ def test_09_config_user_vo_map(self):
...
 
+    def test_10_fix_tomcat_template(self):
+        # Fix EL7 bug in Gratia template
+        if core.el_release() == 7:
+            core.skip_ok_unless_installed(tomcat.pkgname(), 'gratia-service')
+            core.config['gratia.broken_template'] = '/usr/share/gratia/server.xml.template'
+            bad_line = r'\s+sSLImplementation=.*'
+            fixed_line = ' '*15 + 'sslImplementationName="org.glite.security.trustmanager.tomcat.TMSSLImplementation"'
+            files.replace_regexpr(core.config['gratia.broken_template'], bad_line, fixed_line, owner='gratia')
+
```
osg-cat: Have we passed this issue upstream? Have we documented it in the Gratia install guide? Our tests should match our documentation!

brianhlin: Nope because it's very much a 'hacky thing' we do to get our Frankenstein setup to work. I can make a JIRA ticket and assign it to Carl, though

osgtest/library/files.py
brianhlin: I don't see any real use of the input verification here and I think it only causes developer headache: "I want to replace this regular expression in this file with this new line. Wait, the new line has to match the regular expression? What? Why? I want to replace it with whatever I damn well please." That was my response, at least. I've already gone through the rest of the test code to look at usage of this function but since my change lifts the requirement, it doesn't have any affect.

That being said, I could revert it and add the option but I think it's an unnecessary and confusing 'feature' of the function.

osg-cat: Oh, sorry, I did not understand this correctly the first time. The match in question is done with the *replacement* string, not the existing lines. Yeah, that is weird. Who wrote it and do/did they have a good reason?

brianhlin: Cool, github has an option to view files via `git blame`: https://github.com/opensciencegrid/osg-test/blame/master/osgtest/library/files.py#L150-L152

I didn't see @efajardo online at all to ask him about it so I just went with the nuclear option and blew it away because I couldn't think of any good reason to keep it.

efajardo:
@brianhlin I was away. Tues/Wed my reasoning for that was to proctect my self. To make sure what I was replacing had to make sense. I can see how that check would be overkill, so I agree with this change.

PR #28, opened 2016-06-17
Add DummyClass class that ignores everything

matyasselmeci:
Also add `core.init_dummy()` which sets `core.options` and `core._log` to
instances of DummyClass (instead of None).  This lets me import and use
osg-test modules in the interpreter or in standalone scripts, without
getting AttributeErrors because the core functions were trying to do
`_log.write()` and access options.


brianhlin:
I was previously concerned when you changed the initial values of `options` and `_log` but it looks like you solved that with `core.init_dummy()`. LGTM.

PR #32, opened 2016-06-22
get_package_envra: translate missing epochs to '0'

matyasselmeci:
The query run by get_package_envra() returns a missing epoch as the
string '(none)', but an RPM with no epoch is considered the same (by
rpm.labelCompare()) as having the epoch '0'. Translate so the comparison
functions work properly.


brianhlin:
I double-checked the calls to `get_package_envra()` in the rest of osg-test and it doesn't look like any of them use the epoch so this looks good.

PR #34, opened 2016-07-12
Add voms library to set up a voms server without voms-admin

matyasselmeci:
This commit contains two major features:
- VOMS tests have been changed so that most of them will run without
  voms-admin.
- Code for using voms (setting up a vo, adding a user, tearing down the
  vo) has been moved to a separate library named `voms.py`, which can be
  imported and used standalone.

  By default, none of the functions require voms-admin, but two of them,
  `create_vo()` and `add_user()` have a `use_voms_admin` parameter which
  causes them to use the voms-admin-based code instead of the rewritten
  versions. The tests take advantage of this, using voms-admin if it is
  available.

Because many of the test functions have the same requirements (voms
client/server/mysql-plugin, mysql client/server, and, for EL 7, a
minimum version of voms-server), `voms.py` contains `is_installed()` and
`skip_ok_unless_installed()` to verify that the packages to run and use
a voms server are installed.

(SOFTWARE-2346)

### DIFF
--- a/osgtest/library/voms.py
+++ b/osgtest/library/voms.py
@@ -27,7 +27,7 @@ def _get_sqlloc():
     return voms_mysql_so_path
 
 
-def create_vo(vo, dbusername, dbpassword, vomscert, vomskey, use_voms_admin=False):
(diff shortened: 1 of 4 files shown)


matyasselmeci:
http://vdt.cs.wisc.edu/tests/20160708-1542/results.html has successful VMU test results from an earlier version of this branch (contained the earlier version of the edg-mkgridmap fix from #33).

PR #35, opened 2016-07-22
Add tests for GSI-OpenSSH (SOFTWARE-2390)

matyasselmeci:
The tests are one run of gsissh to see if we can connect and run a
simple command, and two runs of gsiscp to see if we can copy files
to/from the remote server.  These tests require gsi-openssh-server,
gsi-openssh-clients, and a proxy.  The code for setting up the file
transfer tests was based off of the gfal2 test code.

A more global change: I changed special_user.py to give the vdttest user
a randomly-generated password.  SSH (including GSISSH) will consider a
user 'disabled' and refuse to log in as that user if the user does not
have a valid encrypted password in /etc/shadow.

### DIFF
--- a/files/test_sequence
+++ b/files/test_sequence
@@ -20,6 +20,7 @@ test_24_tomcat
 test_25_voms_admin
 test_26_bestman
 test_27_osg_info_services
+test_28_gsiopenssh
 test_30_misc
 test_35_osg_configure
 test_38_cvmfs
@@ -41,6 +42,7 @@ test_54_gratia
 test_55_condorce
 test_56_lcgutil
 test_57_osg_info_services
+test_59_gsiopenssh
 test_75_gums
 test_76_tomcat
 test_77_myproxy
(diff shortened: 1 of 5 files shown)


matyasselmeci:
A successful VMU test run of this can be found here: http://vdt.cs.wisc.edu/tests/20160721-1623/packages.html. The definition of the "GSISSH" package set was: `gsi-openssh-server, gsi-openssh-clients, globus-proxy-utils`.

brianhlin:
Since there are changes to `special_user`, could you kick off a VMU run against `osg-tested-internal`?

matyasselmeci:
Submitted.

matyasselmeci:
http://vdt.cs.wisc.edu/tests/20160727-1556/packages.html. Mostly successful. Cleanup error with mysql -- probably unrelated. And a really weird install error on CentOS 6. I don't think I've seen a Python stack trace that long -- and especially not yum. Probably also unrelated.

PR #37, opened 2016-08-26
script rename for gridftp-transfer probe (SOFTWARE-2398)

edquist:
The GridftpTransferProbeDriver script (along with other items from the
gridftp-transfer probe) was replaced gridftp-transfer_meter.

It's not obvious if other changes need to be made also to the tests, but
this much at least clearly will need to be changed...

### DIFF
--- a/osgtest/library/core.py
+++ b/osgtest/library/core.py
@@ -429,6 +429,20 @@ def version_compare(evr1, evr2):
 
     return rpm.labelCompare((epoch1, version1, release1), (epoch2, version2, release2))
 
+def package_version_compare(package_name, evr):
+    """Compare EVR of installed package_name to provided evr and return:
+      -1 if the package version is older than evr,
+       0 if the package version is equal to evr,
+       1 if the package version is newer than evr.
+
+    evr can be a string ("[E:]V[-R]") or 3-element tuple or list.
+
+    This is a wrapper around 'version_compare' that avoids having to call
+    'get_package_envra' and extract (e,v,r)
+    """
+    e,n,v,r,a = get_package_envra(package_name)
+    return version_compare((e,v,r), evr)
+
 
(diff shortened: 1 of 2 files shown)


brianhlin:
1. Did you run automated against your branch and the new `gridftp-transfer` probe?
2. I imagine that this will cause nightlies to fail. @osg-cat, how do you want to deal with this? We could ask Carl to add a version check to his change. Or we could merge this and peg the VMU tests to default to the commit before this one.

matyasselmeci:
I recommend conditionalizing this test on the version of the package which has the change. Some of the functions in the `core` module may help you, such as `get_package_envra` and `version_compare`.

edquist:
Yeah, the current PR will cause the release tests to fail.  I'll update the PR to conditionalize based on the version.

brianhlin:
@edquist: once you've done that, check out step 5i [here](https://twiki.opensciencegrid.org/bin/view/SoftwareTeam/TestRunsAsVMs) and kick off a test run against your branch and osg-development.

edquist:
Thanks @brianhlin and @matyasselmeci !  I've updated the PR to conditionalize it based on the version.

I made new a little wrapper function, `package_version_compare`, which combines `get_package_envra` and `version_compare`, and avoids having to manually break open the envra tuple and reassemble it as (e,v,r).

I kicked off a (very small) VMU run against `edquist:software-2398-gridftp-probe; 3.3; osg-development` and `osg-release` to see if everything's in order.

edquist:
The development tests looked good
http://vdt.cs.wisc.edu/tests/20160826-1409/packages.html

still waiting on the osg (release) tests to come back.

brianhlin:
Release tests also look good http://vdt.cs.wisc.edu/tests/20160826-1457/results.html

edquist:
Hurray for the nightlies!  :)

PR #39, opened 2016-08-31
We don't need osg-configure to run jobs, just the env files

brianhlin:
Just something I noticed when working on the CE tests. Test results look good [here](http://vdt.cs.wisc.edu/tests/20160830-1136/results.html) and I did some spot checks to make sure that the job submission tests ran as expected.


osg-cat:
I don’t really understand the problem that is being addressed, but I guess this is OK.

brianhlin:
If `osg-configure` wasn't installed, job submission tests would fail because the env variables that they were looking for wouldn't be set. It was an unnecessary requirement.

PR #40, opened 2016-08-31
Use hostname for MyProxy tests

matyasselmeci:
In newer versions of Globus libs, certs that do not match the host name
of the service are rejected. Since the host cert for the machine does
not match 'localhost', we need to use the full hostname as the MyProxy
server.

This fixes the MyProxy test failures we saw with globus-gssapi-gsi-12.1.

EL6 test results are here: http://vdt.cs.wisc.edu/tests/20160831-1143/results.html
EL7 test results are pending.


matyasselmeci:
EL7 test results are here: http://vdt.cs.wisc.edu/tests/20160831-1312/packages.html

PR #41, opened 2016-09-07
Improve tomcat7 startup time in VMU tests for non-nightly runs (SOFTWARE-2383)

brianhlin:
[SOFTWARE-2383](https://jira.opensciencegrid.org/browse/SOFTWARE-2383)

### Test results ###

* [With nightly=False (the default)](http://vdt.cs.wisc.edu/tests/20160906-1532/results.html) with the following failures:
  * MyProxy tests failing with 3.3 testing because I didn't grab d1177fc from master
  * Handful of random RHEL7 install and VM startup failures
* [With nightly=True](http://vdt.cs.wisc.edu/tests/20160906-1533/results.html) with the following failures:
  * Same MyProxy and RHEL7 install failures as above
  * Random Bestman/Gfal2 (1x) and EL5 SSL yum failures (4x)

### Summary ###

Added a nightly option that will give us the flexibility to speed up tests in the future for any manual VMU runs so that developers won't have to wait as long for their results. Specifically in this change, we are disabling Tomcat persistence for EL6+ tests since we don't have to bother surviving restarts of the service.

By default, osg-test uses these sped up tests and the nightly test infrastructure will specify that it has a higher tolerance for lengthy test runs.

### DIFF
--- a/osg-test
+++ b/osg-test
@@ -28,6 +28,8 @@ def parse_command_line():
                 "extrarepos": [],
                 "exitonfail": False,
                 "hostcert": False,
+                # Allow VMU nightly tests to have longer test timeouts
+                "nightly": False,
                 "packages": [],
                 "password": 'vdttest',
                 "printtest": True,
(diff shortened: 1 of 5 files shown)


brianhlin:
[Accompanying vm-test-run pull request](https://github.com/opensciencegrid/vm-test-runs/pull/15).

osg-test
```diff
@@ -28,6 +28,8 @@ def parse_command_line():
                 "extrarepos": [],
                 "exitonfail": False,
                 "hostcert": False,
+                # Allow VMU nightly tests to have longer test timeouts
```
osg-cat: Very slight preference to have the comment be on the same line as the element…

PR #43, opened 2016-09-21
Fix condor-cron start/stop to work with systemd

matyasselmeci:
Use systemctl commands for starting/stopping condor-cron. In addition, determine the status of the process by using `systemctl is-active` on EL 7 and `service ... status` on EL 6, instead of looking for a lock file. This fixes errors caused by the removal of the condor-cron init script in SOFTWARE-2439.

The results of a VMU test run are here: http://vdt.cs.wisc.edu/tests/20160921-0857/packages.html

### DIFF
--- a/osgtest/library/service.py
+++ b/osgtest/library/service.py
@@ -115,3 +115,4 @@ def is_running(service_name, init_script=None):
     status, _, _ = core.system(command, 'Checking status of ' + service_name + ' service')
 
     return status == 0
+
--- a/osgtest/library/service.py
+++ b/osgtest/library/service.py
@@ -107,10 +107,11 @@ def is_running(service_name, init_script=None):
     """
     init_script = _init_script_name(service_name, init_script)
 
-    command = ('service', init_script, 'status')
+    if core.el_release() >= 7:
+        command = ('systemctl', 'is-active', init_script)
+    else:
+        command = ('service', init_script, 'status')
+
     status, _, _ = core.system(command, 'Checking status of ' + service_name + ' service')
 
-    if status == 0:
-        return True
-
-    return False
+    return status == 0
--- a/osgtest/tests/test_11_condor_cron.py
+++ b/osgtest/tests/test_11_condor_cron.py
@@ -15,10 +15,15 @@ class TestStartCondorCron(osgunittest.OSGTestCase):
             core.state['condor-cron.running-service'] = True
             self.skip_ok('already running')
 
-        command = ('service', 'condor-cron', 'start')
-        stdout, _, fail = core.check_system(command, 'Start Condor-Cron')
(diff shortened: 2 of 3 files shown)


osgtest/library/service.py
```diff
@@ -107,10 +107,11 @@ def is_running(service_name, init_script=None):
...
+        command = ('systemctl', 'is-active', init_script)
+    else:
+        command = ('service', init_script, 'status')
 
-    if status == 0:
-        return True
+    status, _, _ = core.system(command, 'Checking status of ' + service_name + ' service')
 
-    return False
+    return status == 0
```
brianhlin: Small nitpick, there's no newline at the end of the file

osgtest/tests/test_11_condor_cron.py
```diff
@@ -1,24 +1,31 @@
...
+            command = ('service', 'condor-cron', 'start')
+            stdout, _, fail = core.check_system(command, 'Start Condor-Cron')
+            self.assert_(stdout.find('error') == -1, fail)
+        else:
+            core.check_system(('systemctl', 'start', 'condor-cron'), 'Start Condor-Cron')
+
+        # Give condor-cron time to start up
+        time.sleep(5)
+
+        self.assert_(service.is_running('condor-cron'), "Condor-Cron is not running")
```
brianhlin: `service.is_running()` turns out to work here because the lockfile is the same between the init script and systemd but this isn't always the case (HTCondor-CE comes to mind). If the lockfile names are different, and the init `status()` function checks for the lockfile, it will always fail since the lockfile exists elsewhere. We should update `service.is_running()` to use `systemctl is-active` .

osgtest/tests/test_11_condor_cron.py
```diff
@@ -1,24 +1,31 @@
...
-                     'Condor-Cron run lock file missing')
+        if core.el_release() < 7:
+            command = ('service', 'condor-cron', 'start')
+            stdout, _, fail = core.check_system(command, 'Start Condor-Cron')
+            self.assert_(stdout.find('error') == -1, fail)
+        else:
+            core.check_system(('systemctl', 'start', 'condor-cron'), 'Start Condor-Cron')
+
+        # Give condor-cron time to start up
+        time.sleep(5)
```
brianhlin: This feels hacky perhaps we should add the timeout to `service.is_running()` so we can duck out early if the service comes up within the timeout?

osgtest/tests/test_11_condor_cron.py
matyasselmeci: `service.is_running()` _does_ use is-active. Look at ab4b2f0

brianhlin:
You beat me to it. I noticed that we can't squash merge in this repo :-1:

PR #45, opened 2016-09-28
Fully support systemd in the service library

brianhlin:
This fixes the globus-gatekeeper startup issues in the tests. A rather large set of changes as it moves almost all ([GridFTP has a broken response to `status`](https://jira.opensciencegrid.org/browse/SOFTWARE-2470?filter=12359)) tests to the new service commands; I'd suggest reviewing commit by commit as the files changed are mostly exclusive to and grouped by commits. Let me know if there are any questions at all.

Test results: http://vdt.cs.wisc.edu/tests/20160928-1353/results.html

### DIFF
--- a/osgtest/library/condor.py
+++ b/osgtest/library/condor.py
@@ -20,24 +20,10 @@ def lockfile_path():
     return condor_lockfile
 
 
-def is_running():
-    """True if condor is running, False otherwise
-    On EL5 and EL6, tests the existence of a lockfile. On EL7, runs
-    'service condor status'
-
-    """
-    condor_lockfile = lockfile_path()
-    if condor_lockfile is not None:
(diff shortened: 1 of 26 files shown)


matyasselmeci:
Looks good.
While it's not necessary to do this right now, it might be a good idea to consolidate the repeated pattern of `service.start(...)` followed by `self.assert_(service.is_running(...))` into a single statement, e.g. a `service.checked_start(...)` and `service.checked_stop(...)`

matyasselmeci:
Also, we never use the `init_script` or `sentinel_file` arguments to `service.start()`, so we should drop those and the code that uses those altogether.

matyasselmeci:
I'll merge this as-is - the other comments are for future work.

PR #47, opened 2016-09-29
Temporary fix to handle broken CE 2.0.8 to 2.0.9 updates

brianhlin:
Update tests are broken because `htcondor-ce-2.0.8-2` unconditionally restarts on uninstallation so the CE tests see that the service is already running and subsequent CE tests are skipped. The service restart is a temporary fix (required to reload the `condor_mapfile` to accept our daemons using the generated host cert) to be removed when `htcondor-ce-2.0.9-3` is released so that we don't have a blind spot in the test coverage while it is in testing.

Test results: http://vdt.cs.wisc.edu/tests/20160929-1113/results.html

### DIFF
--- a/osgtest/tests/test_19_condorce.py
+++ b/osgtest/tests/test_19_condorce.py
@@ -94,12 +94,15 @@ gridmapfile -> good | bad
         core.state['condor-ce.schedd-ready'] = False
 
         core.skip_ok_unless_installed('condor', 'htcondor-ce', 'htcondor-ce-client')
-        self.skip_ok_if(service.is_running('condor-ce'), 'already running')
-        service.start('condor-ce')
-
         collector_log, _, _ = core.check_system(('condor_ce_config_val', 'COLLECTOR_LOG'),
                                                 'Failed to query for Condor CE CollectorLog path')
         core.config['condor-ce.collectorlog'] = collector_log.strip().strip()
+
+        if service.is_running('condor-ce'):
+            core.state['condor-ce.schedd-ready'] = True
+            self.skip_ok('already running')
+        service.start('condor-ce')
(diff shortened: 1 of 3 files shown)


osgtest/tests/test_41_jobs.py
```diff
@@ -99,7 +100,7 @@ def test_04_condor_run_pbs(self):
     def test_05_condor_ce_run_condor(self):
         core.skip_ok_unless_installed('htcondor-ce', 'htcondor-ce-client', 'htcondor-ce-condor', 'condor')
 
-        self.skip_bad_unless(core.state['condor-ce.started-service'], 'ce not started')
+        self.skip_bad_unless(service.is_running('condor-ce'), 'ce not started')
```
matyasselmeci: The text should be 'ce not running'

osgtest/tests/test_19_condorce.py
```diff
@@ -99,6 +99,14 @@ def test_05_start_condorce(self):
         core.config['condor-ce.collectorlog'] = collector_log.strip().strip()
 
         if service.is_running('condor-ce'):
+            # Required to accept changes to the mapfile, which caused
+            # issues in the nightly due to bad htcondor-ce-2.0.8-2
+            # packaging remove after OSG 3.3.17. We don't use service.stop()
+            # because it only stops services that we've started
+            command = ('systemctl', 'stop', 'condor-ce')
+            core.check_system(command, 'Stop condor-ce service')
```
matyasselmeci: Won't this fail on EL 6?

brianhlin:
Applied fixes to both. The latter turned out to work because the initial packaging problem was EL7-only.

PR #49, opened 2016-10-20
allow virtual dependencies for torque packages

edquist:
Vaguely SOFTWARE-2484 related.

Apparently some (non-epel, oops) flavors of the torque/pbs packages have
a little different layout, and, for instance, 'torque-mom' is actually a
virtual capability provided by the 'torque-client' rpm.  In that case,
we really only care about 'torque-mom' as a dependency, not necessarily
as a package name.

### DIFF
--- a/osgtest/tests/test_17_pbs.py
+++ b/osgtest/tests/test_17_pbs.py
@@ -32,7 +32,7 @@ set server acl_host_enable = True
 
     def test_01_start_mom(self):
         core.state['pbs_mom.started-service'] = False
-        core.skip_ok_unless_installed(*self.required_rpms)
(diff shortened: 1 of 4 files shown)


edquist:
By the way, I wonder if perhaps `by_dependency=True` should be the default in general, instead of by rpm package names...

brianhlin:
@matyasselmeci got an opinion on turning `by_dependency` on by default?

matyasselmeci:
Fine by me.

brianhlin:
Turning it on by default looks ugly: http://vdt.cs.wisc.edu/tests/20161109-1412/results.html

I'll have to investigate the cause later.

edquist:
Rebased with @brianhlin's merge conflict resolution from https://github.com/edquist/osg-test/pull/3

PR #50, opened 2016-10-25
Refactor systemd support

brianhlin:
@matyasselmeci I've made the changes that you mentioned in #45. Automated test results: http://vdt.cs.wisc.edu/tests/20161024-1711/results.html

### DIFF
--- a/osgtest/tests/test_84_xrootd.py
+++ b/osgtest/tests/test_84_xrootd.py
@@ -13,5 +13,5 @@ class TestStopXrootd(osgunittest.OSGTestCase):
         core.skip_ok_unless_installed('xrootd', by_dependency=True)
         self.skip_ok_if(core.state['xrootd.started-server'] == False, 'did not start server')
 
-        service.stop(core.config['xrootd_service'])
-        self.assert_(not service.is_running(core.config['xrootd_service']), 'XRootD failed to stop')
+        service.check_stop(core.config['xrootd_service'])
+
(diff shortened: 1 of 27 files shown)


osgtest/tests/test_84_xrootd.py
```diff
@@ -12,6 +12,5 @@ def test_01_stop_xrootd(self):
             files.restore('/etc/grid-security/xrd/xrdmapfile', "xrootd")
         core.skip_ok_unless_installed('xrootd', by_dependency=True)
         self.skip_ok_if(core.state['xrootd.started-server'] == False, 'did not start server')
-
-        service.check_stop(core.config['xrootd_service'])
+        service.stop(core.config['xrootd_service'])
```
matyasselmeci: We _do_ package xrootd. How long does stopping it actually take?

brianhlin: It was taking longer than 30 seconds in my previous attempts, I didn't try anything past that.

matyasselmeci:
The last commit bothers me. If we don't wait for the services to stop, then running osg-test twice in a row might contain failures, as the services won't be started up when they're already running, even if they're in the middle of shutting down.

brianhlin:
Fair enough: I tried a few iterations of {{check_stop}} timeouts but couldn't find one that worked and figured we weren't checking before so it didn't hurt to continue to not check it. I'll do some work to try and investigate how long it's actually taking for those services to stop.

bbockelm:
FWIW - I've never seen Xrootd take more than a second or two to stop.  I would be very worried if it doesn't finish after 30s: could indicate a real bug!

brianhlin:
Ah ok, closer investigation shows on EL6, failures were due to:

1. A bad reference to the `tomcat` service name
1. `voms-admin` and `xrootd` services returning incorrect exit codes when checking the status (both return 1, should be 3)
1. `pbs_mom` appears to leave its pid file behind after shutting down

I've added a comment for 1. and will fix 2. after we fix the `voms-admin` and `xrootd` services.

matyasselmeci:
I will merge this. Can you make JIRA tickets for the voms-admin and xrootd fixes?

brianhlin:
xrootd: https://jira.opensciencegrid.org/browse/SOFTWARE-2513
voms-admin: https://jira.opensciencegrid.org/browse/SOFTWARE-2514

PR #53, opened 2016-10-31
Add HTCondor-CE-View test 

brianhlin:
https://jira.opensciencegrid.org/browse/SOFTWARE-2479

Automated testing results here: http://vdt.cs.wisc.edu/tests/20161028-1017/results.html

### DIFF
--- a/osgtest/tests/test_55_condorce.py
+++ b/osgtest/tests/test_55_condorce.py
@@ -3,7 +3,7 @@
 
 import os
 import re
-import socket
+import urllib2
 
 import osgtest.library.condor as condor
 import osgtest.library.core as core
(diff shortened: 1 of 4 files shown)


osgtest/tests/test_55_condorce.py
```diff
@@ -137,23 +138,33 @@ def test_06_use_gums_auth(self):
...
         self.assert_(re.search(r'Authorized:\s*TRUE', stdout), 'could not authorize with GSI')
+
+    def test_07_ceview(self):
+        core.config['condor-ce.view-listening'] = False
+        self.general_requirements()
+        core.skip_ok_unless_installed('htcondor-ce-view')
+        view_sock = socket.socket()
+        try:
+            try:
+                view_sock.connect((core.get_hostname(), int(core.config['condor-ce.view-port'])))
```
djw8605: Rather than just connecting, why don't you do a urllib GET?

brianhlin: I think my initial reasoning was to just do the absolute basic check to see that the CE View was listening on the expected port. Using `urllib` seems reasonable and we can do a basic assertion of the contents to ensure that it is indeed the CE VIew that's listening on the port. I'll make a quick commit and test it.

brianhlin:
New automated tests here: http://vdt.cs.wisc.edu/tests/20161031-1517/results.html

matyasselmeci:
LGTM.

PR #56, opened 2016-11-08
Use service library for gridftp and add badskips if it can't be started

matyasselmeci:
This requires the version of gridftp that was just released today, because of the init script exit status fix.

Test results are here: http://vdt.cs.wisc.edu/tests/20161108-1537/packages.html


osg-cat:
I don’t understand the comment — osg-test needs to work for past versions of GridFTP.

PR #58, opened 2016-11-22
Use TrueType fonts, which are properly scaled on HiDPI displays

matyasselmeci:


brianhlin:
It looks like the tkFonts library is provided by the `tkinter` package so we'll need to add this requirement. We can separate the `osg-test-log-viewer` into its own subpackage. 

Otherwise LGTM

PR #59, opened 2016-11-30
catalina.CYMD.log dates are in UTC

edquist:
As @timtheisen explained, tomcat startup tests fail for jobs started
after 6pm CST, as the date in the catalina log file name is UTC, not
localtime.

Note that this fixes the issue with tests started between midnight UTC
(ie, 6pm CST) and midnight localtime, but does not handle the case where
a test starts just before midnight UTC, but the sentinel token is
written afterwards to a new log file for the next UTC date.

### DIFF
--- a/osgtest/library/tomcat.py
+++ b/osgtest/library/tomcat.py
@@ -1,5 +1,5 @@
 import os
-from datetime import date
+import time
 
 import osgtest.library.core as core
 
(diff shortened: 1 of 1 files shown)


brianhlin:
@edquist could you kick off full set of manual tests (now) and one post-6pm to see how everything works? I think this should work ok for the set of tests that have Gratia involved because that's what causes the GMT rollover. However, when Gratia isn't involved, I think the Catalina log will be named using the local time date and we'd miss it in that 6pm-midnight CT span.

edquist:
Sounds good, will do.

brianhlin:
@edquist Unfortunately, the error patterns of the [post-6pm run](http://vdt.cs.wisc.edu/tests/20161130-1800/results.html) are what I expected above (except for a single test run that started after 2am). I can think of a real quick, dumb fix: use `core.monitor()` to watch the N-dated log, where N is the local time date,  and if that fails to find the sentinel, search for the sentinel in the N+1-dated log.

edquist:
Yeah it was too good to be true.  Just falling back to the next dated filename if the first monitor fails would work.  (Unless we wanted to allow monitoring for multiple files at once.... mmm :)

PR #60, opened 2016-11-30
Fix pbs server start

brianhlin:
Test results here: http://vdt.cs.wisc.edu/tests/20161129-0935/results.html

The PBS server startup test started being "skipped" when we replaced the `core.system()` call to stop the service with `service.stop()`*, which doesn't start the service unless we set `core.state['pbs_server.started-service'] = True`**. We didn't have a complete PBS blind spot in the tests because the `condor_ce_trace_pbs` jobs seemed to run fine...

\* We stopped the service because we had to run `service pbs create` to create the initial config and it did a bad job of stopping the service afterwards, which we needed to do to drop additional config (maybe?).

** We really shouldn't be setting `core.state` variables in our library functions, at the very least, not the `<service name>.started-service` keys because it makes following service state confusing.

### DIFF
--- a/osgtest/tests/test_41_jobs.py
+++ b/osgtest/tests/test_41_jobs.py
@@ -63,11 +63,8 @@ class TestRunJobs(osgunittest.OSGTestCase):
                                       'torque-mom', 'torque-server', 'torque-scheduler')
         self.skip_bad_unless(core.state['globus-gatekeeper.running'], 'gatekeeper not running')
         self.skip_bad_unless(core.state['jobs.env-set'], 'job environment not set')
-        if (not core.state['torque.pbs-configured'] or
-            not core.state['torque.pbs-mom-running'] or
-            not core.state['torque.pbs-server-running'] or
-            not core.state['globus.pbs_configured']):
-            self.skip_bad('pbs not running or configured')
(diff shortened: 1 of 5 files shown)


osgtest/tests/test_41_jobs.py
```diff
@@ -95,6 +95,7 @@ def test_05_condor_ce_run_condor(self):
         core.skip_ok_unless_installed('htcondor-ce', 'htcondor-ce-client', 'htcondor-ce-condor', 'condor')
 
         self.skip_bad_unless(service.is_running('condor-ce'), 'ce not running')
+        self.skip_bad_unless(service.is_running('condor'), 'ce not running')
```
brianhlin: Oops, kill-yank is the devil.

edquist:
Yay... Looks like this simplifies the PBS server logic a lot.

Just to make a note of it here: As alluded to in the PR comments, with these changes we are now doing the sort-of-hacky replacement for `service pbs_server create` on all platforms (instead of just on EL7), because the `create` function in the `pbs_server` init.d script was causing problematic side effects for our tests.

There is kind of a lot that's changed here, but the successful VMU test run gives some additional confidence that things are now working as expected.

PR #62, opened 2016-12-15
Handle different alternatives formats

brianhlin:
Previous formatting of java alternatives:

```
*+ 1           /usr/lib/jvm/java-1.7.0-openjdk-1.7.0.111-2.6.7.2.el7_2.x86_64/jre/bin/java
```

CentOS/RHEL 7 appear to have changed their java alternatives format to the following:

```
*+ 1           java-1.8.0-openjdk.x86_64 (/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.111-2.b15.el7_3.x86_64/jre/bin/java)
   2           java-1.7.0-openjdk.x86_64 (/usr/lib/jvm/java-1.7.0-openjdk-1.7.0.121-2.6.8.0.el7_3.x86_64/jre/bin/java)
```
Here's a run with failing java selection (CentOS and RHEL 7 BeStMan and GUMS tests) : http://vdt.cs.wisc.edu/tests/20161212-1621/results.html
Here's a run with the fix (I excluded 'All*' tests with the hope they'd finish earlier): http://vdt.cs.wisc.edu/tests/20161213-1542/results.html

I also removed the unused libraries from the java tests since I noticed that while I was in there.

### DIFF
--- a/osgtest/library/java.py
+++ b/osgtest/library/java.py
@@ -26,9 +26,9 @@ def select_ver(java_type, version):
     """Select the specified version of java in the alternatives"""
     alternatives = _run_alternatives(java_type, '\n', 'find %s version for selection' % java_type)
     selection = ''
-    for alt in re.finditer(r'(\d+)\s+(\/.*\n)', alternatives):
-        if version in alt.group(2):
-            selection = alt.group(1)
+    for alt in alternatives.split('\n'):
+        if version in alt:
+            selection = re.match(r'[\s\*\+]*(\d+)\s+', alt).group(1)
     _run_alternatives(java_type, '%s\n' % selection, 'select %s version' % java_type)
 
 def get_ver(java_type):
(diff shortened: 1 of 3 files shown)


osgtest/library/java.py
```diff
@@ -26,22 +26,22 @@ def select_ver(java_type, version):
     """Select the specified version of java in the alternatives"""
     alternatives = _run_alternatives(java_type, '\n', 'find %s version for selection' % java_type)
     selection = ''
-    for alt in re.finditer(r'(\d+)\s+(\/.*\n)', alternatives):
-        if version in alt.group(2):
-            selection = alt.group(1)
+    for alt in alternatives.split('\n'):
+        if version in alt:
+            selection = re.match(r'[\s\*\+]*(\d+)\s+', alt).group(1)
```
osg-cat: On EL 6, is it still correct to match and use the path starting with “/usr/lib”? And on EL 7, you need to use the package name-like thing?

brianhlin: It doesn't appear to be a split between EL6/EL7 since SL7 did not fail in the same way.

osg-cat: Well, historically Scientific Linux gets changes after RHEL and CentOS, so perhaps SL 7 will get this change, too, at some point. In any case, my underlying question was whether the new code works for both cases correctly.

brianhlin: Yes, it does work in both cases. I did however, make an unnecessary change to `get_ver()`, not sure what I was thinking yesterday, so I'll revert it and rebase my branch.

brianhlin:
Test results after reverting the `get_ver()` change look good: http://vdt.cs.wisc.edu/tests/20161215-1018/results.html

brianhlin:
Can I go ahead and merge this?

osg-cat:
That is fine with me.

PR #63, opened 2016-12-15
Test HTCondor-CE + Slurm interaction in automated tests (SOFTWARE-2421)

brianhlin:
Test results [here]( http://vdt.cs.wisc.edu/tests/20161214-1435/results.html); failures, although festive, are unrelated.

This set of commits:

* Separates out the munge service setup since it's required by both PBS and SLURM tests
* Installs slurm from osg-contrib if we're running 'All*' tests
* Fixes a mysql.is_running bug and PBS startup bug that I noticed when testing
* Added get_stat() function to handle missing files. This was duplicated throughout the tests before the function.

@osg-cat: I don't think this needs much of a review, assign as you see fit.

### DIFF
--- a/osgtest/library/core.py
+++ b/osgtest/library/core.py
@@ -55,6 +55,11 @@ _log_filename = None
 _last_log_had_output = True
 _el_release = None
 
+SLURM_PACKAGES = ['slurm',
+                  'slurm-munge',
+                  'slurm-perlapi',
+                  'slurm-plugins',
+                  'slurm-sql']
 
 # ------------------------------------------------------------------------------
 # Global Functions
--- a/osgtest/library/core.py
+++ b/osgtest/library/core.py
@@ -1,5 +1,6 @@
 """Support and convenience functions for tests."""
 
+import errno
 import os
 import os.path
 import pwd
@@ -139,6 +140,14 @@ def remove_log():
     """Removes the detailed log file; not for general use."""
     os.remove(_log_filename)
 
+def get_stat(filename):
+    '''Return stat for 'filename', None if the file does not exist'''
+    try:
+        return os.stat(filename)
+    except OSError, exc:
+        if exc.errno == errno.ENOENT:
+            return None
+        raise
 
 def monitor_file(filename, old_stat, sentinel, timeout):
     """Monitors a file for the sentinel regex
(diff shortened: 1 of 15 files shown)


osgtest/library/core.py
```diff
@@ -134,6 +140,13 @@ def remove_log():
     """Removes the detailed log file; not for general use."""
     os.remove(_log_filename)
 
+def get_stat(filename):
+    '''Return stat for 'filename', None if the file does not exist'''
+    try:
+        return os.stat(filename)
+    except OSError, exc:
+        if exc.errno == errno.ENOENT:
```
osg-cat: OK, so if the file does not exist, you return None. What if there is a different error? Then you still return None, IIRC?

osgtest/library/mysql.py
```diff
@@ -32,7 +32,7 @@ def stop():
     service.check_stop(daemon_name())
 
 def is_running():
-    service.is_running(daemon_name())
+    return service.is_running(daemon_name())
```
osg-cat: Clearer, perhaps, but not different in functionality, I think.

osgtest/tests/test_29_slurm.py
```diff
@@ -0,0 +1,107 @@
...
+PartitionName=debug Nodes=%(short_hostname)s Default=YES MaxTime=INFINITE State=UP
+ReturnToService=2
+SlurmUser=slurm
+SlurmctldDebug=debug5
+SlurmctldTimeout=300
+SlurmctldLogFile=%(ctld_log)s
+SlurmdLogFile=/var/log/slurm/slurm.log
+SlurmdDebug=debug5
+StateSaveLocation=/var/spool/slurmd
+"""
```
osg-cat: Don’t we have a canonical location in which to place files?

osgtest/library/mysql.py
brianhlin: Actually, we need to explicitly return:

```
def foo():
...     return True
... 
>>> def bar():
...     foo()
... 
>>> baz=bar()
>>> baz
>>>
>>> def bar():
...     return foo()
... 
>>> baz=bar()
>>> baz
True
```

osgtest/library/core.py
brianhlin: Good catch, you're right. My intention was to `raise` for all other non-`OSError`s. I'll fix the commit.

brianhlin: It's fixed now. Any other comments or am I free to merge?

PR #66, opened 2017-01-17
Maintain order of package list specified by param files

brianhlin:
Python docs claim that `OrderedDict` was added in Python 2.7 but it appears to be available in python core on all the Python 2.6 installations I've looked at (osghost, fermicloud VMs, my laptop, Carl's CSL machine).

@edquist: I added you as a requested reviewer since we were talking about `OrderedDict`.

### DIFF
--- a/osgtest/tests/special_install.py
+++ b/osgtest/tests/special_install.py
@@ -32,7 +32,7 @@ class TestInstall(osgunittest.OSGTestCase):
 
         # FIXME: Install slurm out of contrib if we're running 'All' tests until
         # SOFTWARE-1733 gives us a generalized solution
-        if any(x in pkg_repo_dict.keys() for x in ['osg-tested-internal', 'osg-tested-internal-gram']) and \
+        if any(x in pkg_repo_dict for x in ['osg-tested-internal', 'osg-tested-internal-gram']) and \
            float(core.osg_release()) > 3.2:
             all_slurm_packages = core.SLURM_PACKAGES + ['slurm-slurmdbd']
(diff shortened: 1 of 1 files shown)


edquist:
Ok, so, as I read it the OrderedDict thing does fix address the goal of this PR.  Feel free to merge.

Note though that the only place the OrderedDict (or, rather, a dictionary at all) is relevant is in [this slurm packages section](https://github.com/brianhlin/osg-test/blob/84e4c1e7308f6ac21c6bc6833ccf889c4cf3f15c/osgtest/tests/special_install.py#L35-L38) (without that, just a simple `list` could be used.)

Also note that in that section, `core.options.extrarepos` gets replaced with `['osg-contrib']` for the slurm packages, rather than added to the list of extra repos.  Brian assures me that this doesn't matter  ;)

Also, not specific to this PR but related to that section, [here](https://github.com/brianhlin/osg-test/blob/84e4c1e7308f6ac21c6bc6833ccf889c4cf3f15c/osgtest/tests/special_install.py#L35), the membership test does not actually take advantage of the fact that a dictionary is used because `pkg_repo_dict.keys()` returns a `list` object ... to use the dictionary to test membership you actually want to test just the condition `x in pkg_repo_dict`.

brianhlin:
Thanks, as always, for the thorough review. I've fixed the membership check. 

As for `core.options.extrarepos` getting replaced with `['osg-contrib']` for slurm packages, the only case where it would matter would be if a user wanted to install slurm through a separate repository. I don't think this is a use case we really expect to test in the nightlies and if a user wanted to test this manually, they could pre-install the slurm packages they're interested in.

PR #67, opened 2017-01-18
Limit condor_{ce_,}run tests to 10 minutes

brianhlin:
@edquist: Requesting that you look this over since you wrote the `core.__run_command()` timeout stuff. Should be a short review. 

### Background 

`condor_*run` will patiently wait for jobs to complete even if the job has no chance of completing. Utilize Carl's timeout option to `core.check_system()` to avoid TIMEOUT failures.

### Test Results

Before this commit: http://vdt.cs.wisc.edu/tests/20170109-1206/results.html
After this commit: http://vdt.cs.wisc.edu/tests/20170118-1100/results.html


edquist:
Looks fine to me.

PR #72, opened 2017-03-16
Travis-CI support (SOFTWARE-2313)

brianhlin:
Since we're limited to 5 concurrent builds and we're kicking so many off, we should make sure that we set the opensciencegrid/osg-test travis-ci limit to 2 builds at a time.

### DIFF
--- a/travis-ci/test_inside_docker.sh
+++ b/travis-ci/test_inside_docker.sh
@@ -1,0 +1,54 @@
+#!/bin/sh -xe
+
+OS_VERSION=$1
+PACKAGE=$2
+
+ls -l /home
+
+# Clean the yum cache
+yum -y clean all
+yum -y clean expire-cache
+
+# First, install all the needed packages.
+rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-${OS_VERSION}.noarch.rpm
+
+# Broken mirror?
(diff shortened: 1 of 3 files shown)


djw8605:
Why the limit?  I think we should just let it go, full speed!

brianhlin:
We'd potentially be starving other opensciencegrid Travis-CI builds and since the osg-test builds take a while to run, I'd rather have them just run a little longer.

djw8605:
Ok, seems reasonable.

djw8605:
Why didn't Travis-CI build this pull request?

brianhlin:
Maybe because I set up the github/Travis-CI integration after submitting the PR? If it's helpful, here are the results from my fork:

https://travis-ci.org/brianhlin/osg-test

travis-ci/test_inside_docker.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Clean the yum cache
+yum -y clean all
+yum -y clean expire-cache
+
+# First, install all the needed packages.
+rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-${OS_VERSION}.noarch.rpm
+
+# Broken mirror?
+echo "exclude=mirror.beyondhosting.net" >> /etc/yum/pluginconf.d/fastestmirror.conf
```
djw8605: I'm curious if this is still needed.

djw8605:
Ah great, now it's building!  I will approve when the build succeeds.

PR #73, opened 2017-03-30
Don't depend on htcondor-ce-condor

brianhlin:
`htcondor-ce-condor` just provides config for `htcondor-ce` and the `condor` backend. Installing `htcondor-ce` brings all the actual packages necessary for testing CE -> condor job submission; we already throw down half of the necessary config by writing our own job routes, so we just need to add the backend `condor` configuration


matyasselmeci:
Why? Is this what we tell our users?

brianhlin:
Well if we want to go by what we tell our users, we should be looking for the `osg-ce-<batch system>` package, which requires the `htcondor-ce-<batch system>`. The underlying packages and half the config are already there, I don't see a point in skipping CE + condor testing in this case.

matyasselmeci:
OK.

PR #76, opened 2017-05-18
Provide runtime JSP implementation for Tiles

brianhlin:
Necessary for SOFTWARE-2652


brianhlin:
Alright, after actually looking at what `files.replace_regexpr()` was doing, I fixed it. Here's a set of tests to verify that it doesn't break tomcat elsewhere: http://vdt.cs.wisc.edu/tests/20170519-1303/results.html

Awaiting test results against the new VOMS Admin server before declaring complete success.

brianhlin:
The latest commits work around the issues with the gratia configuration overwriting `server.xml` by manually editing the file after the configuration if `gratia-service` is installed. Test results [here](http://vdt.cs.wisc.edu/tests/20170523-1207/results.html) look good.

PR #80, opened 2017-06-12
LCMAPS VOMS plugin tests

brianhlin:
This adds testing of the LCMAPS VOMS plugin to the nightlies, replacing the gridmap auth we were testing before. Other things to note:

* The suite still tests `edg-mkgridmap`and `glexec` individually.
* Force GSI auth in HTCondor-CE to ensure that the LCMAPS auth gets exercised. Before it would just fallback to FS auth.
* I'm not sure if XRootd + LCMAPS auth gets exercised (my guess is it doesn't) but added the env variables anyway

### DIFF
--- a/osgtest/tests/test_55_condorce.py
+++ b/osgtest/tests/test_55_condorce.py
@@ -158,8 +158,8 @@ gums.authz=https://%s:8443/gums/services/GUMSXACMLAuthorizationServicePort
 ''' % (hostname, hostname)
 
         core.config['condor-ce.lcmapsdb'] = os.path.join('/', 'etc', 'lcmaps.db')
-        core.config['condor-ce.gums-properties'] = '/etc/gums/gums-client.properties'
-        core.config['condor-ce.gsi-authz'] = '/etc/grid-security/gsi-authz.conf'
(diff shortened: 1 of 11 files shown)


brianhlin:
Test results here:

http://vdt.cs.wisc.edu/tests/20170611-1818/results.html
http://vdt.cs.wisc.edu/tests/20170612-0909/results.html

osgtest/tests/test_55_condorce.py
```diff
@@ -150,6 +150,7 @@ def test_07_ping_with_gums(self):
 gums.authz=https://%s:8443/gums/services/GUMSXACMLAuthorizationServicePort
 ''' % (hostname, hostname)
 
+        core.config['condor-ce.lcmapsdb'] = os.path.join('/', 'etc', 'lcmaps.db')
         core.config['condor-ce.gums-properties'] = '/etc/gums/gums-client.properties'
```
edquist: Minor: in the same block here, paths are just written out, eg, `/etc/lcmaps.db` instead of `os.path.join('/', 'etc', 'lcmaps.db')`

PR #82, opened 2017-06-28
Drop unused tests (SOFTWARE-2751)

brianhlin:
**Test results**

http://vdt.cs.wisc.edu/tests/20170622-1450/results.html
http://vdt.cs.wisc.edu/tests/20170622-1653/results.html

Also bumping the version number


matyasselmeci:
Yessss!!!!

PR #83, opened 2017-07-06
Fix SELinux issues with gsi-openssh tests

matyasselmeci:
On EL7, SELinux forbids sshd (and gsisshd) from opening unregistered ports (e.g. 2222). This fix allows gsisshd to use port 2222 for the duration of the osg-test run. Some contortions were necessary so that the tests OkSkip if we need policycoreutils-python and can't find it.


matyasselmeci:
Test results: http://vdt.cs.wisc.edu/tests/20170706-1718/results.html

PR #87, opened 2017-07-24
Run 3.4 Travis-CI tests by default

brianhlin:
Also, fix location of CE `condor_mapfile` if we don't generate a hostcert


bbockelm:
How did we survive before without the `--privileged` option?  Did we tickle some new case for OSG 3.4?

brianhlin:
`--privileged` was used for EL7 test but not EL6. I was seeing cgroup errors in the procd for HTCondor so I'm guessing that cgroups were enabled by default in 8.6.

bbockelm:
Fair enough.  Could you put a small comment in the code along those lines?  Just so future readers can clearly understand what the privileged mode is for.

PR #91, opened 2017-08-10
Limit status checks before starting services (SOFTWARE-2846)

brianhlin:
Plus bonus commit for TODO item that wasn't marked as such.

Test results here: http://vdt.cs.wisc.edu/tests/20170810-0913/results.html


brianhlin:
@matyasselmeci how's it look now?

PR #96, opened 2017-09-08
condor-ce: restore lcmaps.db and other auth files immediately after t…

matyasselmeci:
…est_07_ping_with_gums

That test should _temporarily_ modify lcmaps.db and a
few other auth files in order to test that gums auth
works with condor-ce. However, it currently restores
them only after other tests have run -- tests which may
depend on lcmaps.db, e.g. gsiopenssh.


matyasselmeci:
http://vdt.cs.wisc.edu/tests/20170908-1436/packages.html

PR #99, opened 2017-09-27
Move test_58_gfal2util to its correct position in the test sequence

matyasselmeci:


matyasselmeci:
See test run here: http://vdt.cs.wisc.edu/tests/20170926-1733/packages.html

matyasselmeci:
Yeah, the tests are being skipped because they depend on bestman.

PR #100, opened 2017-09-27
Remove GRAM code

matyasselmeci:
Test results: http://vdt.cs.wisc.edu/tests/20170926-2001/packages.html


brianhlin:
Looking through the results, what's with osg-configure unit tests still having `TestGram` tests?

matyasselmeci:
Those osg-configure tests make sure that trying to enable gram results in an error.

PR #101, opened 2017-09-27
Use HAVEGED to add entropy (SOFTWARE-2880)

matyasselmeci:
http://vdt.cs.wisc.edu/tests/20170927-1122/packages.html


files/test_sequence
```diff
@@ -64,3 +65,4 @@ test_94_munge
 test_95_mysqld
 test_96_proxy
 test_97_java
+test_98_haveged
```
brianhlin: Could you add a newline?

PR #103, opened 2017-09-28
Various misc cleanup tasks

matyasselmeci:
http://vdt.cs.wisc.edu/tests/20170928-1321/packages.html

### DIFF
--- a/osgtest/library/voms.py
+++ b/osgtest/library/voms.py
@@ -148,6 +148,7 @@ def is_installed():
         if not core.dependency_is_installed(dep):
             return False
 
+    # TODO: drop this check when 3.3 is completely EOL
     if core.el_release() >= 7:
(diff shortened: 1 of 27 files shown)


osgtest/library/voms.py
```diff
@@ -148,6 +148,7 @@ def is_installed():
         if not core.dependency_is_installed(dep):
             return False
 
+    # TODO: drop this check when 3.3 is completely EOL
```
brianhlin: We can drop this entire library and all corresponding tests once 3.3 is EOL

matyasselmeci: We won't be testing VOMS at all? Not just voms-admin?

brianhlin: Yea, not really. Maybe `voms-proxy-fake` but we don't need the library for it or really any of what happens in `test_50_voms.py`

PR #107, opened 2017-11-21
Grab slurm from development repos instead of contrib (SOFTWARE-2994)

brianhlin:
Test results are good: http://vdt.cs.wisc.edu/tests/20171117-1442/results.html


brianhlin:
Oops, accidentally deleted the origin branch

PR #111, opened 2018-01-12
decorators to tests to osg release series (SOFTWARE-2759)

edquist:
this allows test classes to add decorators like:
```
    @osgrelease(3.3)
    def test_foo_33_only(self):
        ...
```
or
```
    @osgrelease(3.3, 3.4)
    def test_foo_33_or_34(self):
        ...
```
to restrict running a particular test function to when the currently installed osg release series is in the provided list.

`@osgrelease(3.3)` is added for the bestman tests, as an example, but the other `skip_ok_unless_installed()`'s still need to be reviewed for 3.3-only packages.

### DIFF
--- a/osg-test-log-viewer
+++ b/osg-test-log-viewer
@@ -44,7 +44,7 @@ def load_logdata_from_handle(handle):
                                     line)
         match = (re.match(r'message:\s*%s:\s*(.+?)\s*$' % datetime_re, line) or
                  command_match or
-                 re.match(r'((?:FAIL|ERROR|BAD SKIPS|OK SKIPS).+?)\s*$', line) or
+                 re.match(r'((?:FAIL|ERROR|BAD SKIPS|OK SKIPS|EXCLUDED).+?)\s*$', line) or
                  re.match(r'(OSG-TEST LOG)', line))
         output_delimiter = re.match(r'(?:STDOUT|STDERR):[{}]', line)
 
--- a/osgtest/library/core.py
+++ b/osgtest/library/core.py
@@ -715,3 +715,23 @@ def remove_cert(target_key):
         target_dir = state[target_key + '-dir']
         if len(os.listdir(target_dir)) == 0:
             os.rmdir(target_dir)
+
+def osgrelease(*releases):
+    """
+    Return a decorator that will only call its function when the current
+    osg_release version is specified in the list of releases.
+
+        class TestFoo(osgunittest.OSGTestCase):
+
+            @osgrelease(3.4)
+            def test_bar_34_only(self):
+                ...
+    """
+    releases = map(str, releases)  # convert float args to str
+    def osg_release_decorator(fn):
+        def run_fn_if_osg_release_ok(*args, **kwargs):
+            if osg_release() in releases:
+                return fn(*args, **kwargs)
+        return run_fn_if_osg_release_ok
+    return osg_release_decorator
+
--- a/osgtest/library/core.py
+++ b/osgtest/library/core.py
@@ -348,17 +348,8 @@ def skip_ok_unless_installed(*packages_or_dependencies, **kwargs):
     if isinstance(packages_or_dependencies[0], (list, tuple)):
         packages_or_dependencies = packages_or_dependencies[0]
 
-    missing = []
-    if by_dependency:
-        dependencies = packages_or_dependencies
-        for dependency in dependencies:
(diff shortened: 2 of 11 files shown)


brianhlin:
Looking at the CI, it looks like you'll need to decorate all of the tests inside `test_52_bestman`

edquist:
Yeah, which is a bit confusing to me because those tests did not have calls to `skip_ok_unless_installed()` for the bestman2 packages...

I've added `@osgrelease(3.3)` decorators to the rest of the `test_52_bestman` tests, but I wonder if this is really the way we want to handle the issue in SOFTWARE-2759.  From yesterday:

http://vdt.cs.wisc.edu/tests/20180111-1933/results.html

As you can see, the `@osgrelease(3.3)` tests still show up in the list, only as `ok` instead of `okskip`.  Perhaps what we really want is a way to exclude the unwanted tests from the list to run, rather than reporting them as `ok` instead of `okskip`?

matyasselmeci:
They should definitely not be `ok`. I'm 50-50 on dropping them from the list vs. using `okskip`, but perhaps it would be even better to make a new `excluded` state or something.

brianhlin:
@matyasselmeci the whole point of this was that the 3.3 only tests were causing a lot of noise in the `okskip` numbers of 3.4. Given the choice between dropping them from the list vs a separate sate, I'd prefer an `excluded` (or `deprecated`?) state.

brianhlin:
@matyasselmeci good point, `excluded` works.

@edquist that's fine, we don't guarantee that 3.3-only packages will work in a 3.4 environment

matyasselmeci:
We don't have to add the decorators for everything; if, say existing GUMS installations are still supported in 3.4, then we still want to test them.

matyasselmeci:
_New_ GUMS installations aren't, but we did go through some effort to make sure that 3.4 doesn't break existing GUMS installations.

brianhlin:
Alright, so two separate things here:

1. I agree in principle, we don't need to include decorators on everything. If for some reason we need to support a package in version N that is only available in version N-1, we just won't apply the decorator. 
1. Yes, we put in that effort as a courtesy to our users because of the short notice of the retirement. However, if a user comes to us with a 3.3 GUMS and a 3.4 CE/gridftp server/etc, our recommendation to them should be an LCMAPS VOMS transition.

So @edquist the takeaway from this conversation is that we want decorated tests that are not run to be added to an `excluded` list

edquist:
@brianhlin, the above commit 344e5b6 creates a separate `excluded` category/list for the decorated tests run on the wrong osg series.  I'm not sure what the travis-ci failure is about.

Here's a VMU run with the above:
http://vdt.cs.wisc.edu/tests/20180116-1233/packages.html

The list of tests at the top looks fine for `excluded` tests, but the detailed `EXCLUDED` section below (after `OK SKIPS`) for some reason shows the python traceback instead of just the oneline summary as seen in the `OK SKIPS` section.  Not sure what I missed there but I'm guessing we don't want it quite as-is yet.

brianhlin:
@edquist it looks like there were some issues with the Travis VMs. I restarted them but note that you should be able to do that yourself next time by going to the test details and logging in.

edquist:
Think I fixed the issue with the `EXCLUDED` section.  Kicking off another VMU run for it...

edquist:
Yup, the `EXCLUDED` section is looking better now:
http://vdt.cs.wisc.edu/tests/20180122-1901/packages.html

@brianhlin, I think this is ready to go.

brianhlin:
LGTM. I've got a commit to exclude GUMS/edg-mkgridmap tests incoming so I won't merge until that looks good.

edquist:
@brianhlin, in order to use the new `osgrelease(...)` function for a decorator without fully-qualifying it, you need to:
```from osgtest.library.core import osgrelease```

I can make a fix commit for that, or else I can reset my branch to 2c9433e and you can make more commits to use the decorators in separate PRs.

brianhlin:
To what end? Since we were already importing `core`, I decided to just use `core.osgrelease`.

edquist:
Derp, nevermind.  I misread and imagined you were doing just `@osgrelease(3.3)`.  Yeah, `@core.osgrelease(3.3)` (as you have it) works just fine.

brianhlin:
That's why it passes CI :)

PR #113, opened 2018-01-26
SOFTWARE-2911

efajardo:
As part of [SOFTWARE-2911](https://jira.opensciencegrid.org/browse/SOFTWARE-2911). I added the decorators to the bestman only for 3.3. But also I added some more tests with gfal. Finally since I was already there I picked of @edquist work and added the decorator for Major OS versions to exclude.

The [VMU tests ](http://vdt.cs.wisc.edu/tests/20180125-1905/results.html) are green.

### DIFF
--- a/osgtest/library/core.py
+++ b/osgtest/library/core.py
@@ -731,26 +731,26 @@ def osgrelease(*releases):
         return run_fn_if_osg_release_ok
     return osg_release_decorator
 
-def osg_os_release(*releases):
+def elrelease(*releases):
     """
     Return a decorator that will only call its function when the current
-    osg_release version is specified in the list of releases; otherwise
+    el_release version is specified in the list of releases; otherwise
     ExcludedException is raised and the test is added to the 'excluded'
     list.
 
         class TestFoo(osgunittest.OSGTestCase):
 
-            @osgrelease(3.4)
-            def test_bar_34_only(self):
+            @elrelease(7)
+            def test_bar_el7_only(self):
                 ...
     """
-    releases = map(str, releases)  # convert float args to str                                                                                           
-    def osg_os_release_decorator(fn):
-        def run_fn_if_osg_os_release_ok(*args, **kwargs):
+    def el_release_decorator(fn):
+        def run_fn_if_el_release_ok(*args, **kwargs):
             if el_release() in releases:
                 return fn(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


edquist:
@efajardo, i pushed an update for your os release decorator --

- casting releases to str is not the right thing here: they are ints
- name should reflect el release, not based on osg release
- update comment example

> decorator for Major OS versions to exclude
Just to make sure we're on the same page here: these decorators specify the only release versions to _include_, and all others are excluded.  (That is, the ones specified are the ones _not_ excluded.)

That was your intention, to only run those test for EL7, right?

edquist:
(Updating the PR title so that it shows up automatically in the JIRA ticket.)

efajardo:
Wow @edquist thanks for fixing it. Yes that was my intention to only have it run in the specified OS major release version

edquist:
Cool.  In that case the changes look good.

I'm kicking off another VMU run using this branch to make sure everything goes ok, and then we can merge this.

If you look at the [previous run](http://vdt.cs.wisc.edu/tests/20180125-1905/results.html), you'll see that even the EL7 tests show the new tests (like `test_05_copy_local_to_server_gfal2_util`) as `excluded` -- so they never ran at all.

This is because in the [os release decorator](https://github.com/opensciencegrid/osg-test/pull/113/commits/8134721186bd9ebed1cdecc6f87b043c8127fe9c#diff-838184dd6ca065ba35252cc39c9d2cfeR747), `releases` were mapped to `str`s, which never compare as equal to the value of `el_release()`, which is always an `int`.  (The osg releases were mapped to `str` because that's what `osg_release()` returns.) ... Anyway, it should be fixed in 314ddc0.

edquist:
By the way, in case it's not clear, it should be possible to combine these two types of decorators now, if we ever want to restrict both the OSG release and the EL release

Eg:
```
    @core.osgrelease(3.4)
    @core.elrelease(7) 
    def test_123_foo_osg34_el7_only(self):
        # ...
```

edquist:
All green:

http://vdt.cs.wisc.edu/tests/20180126-1740/packages.html

I verified that `test_05_copy_local_to_server_gfal2_util` runs for EL7 and is excluded for EL6.

efajardo:
THanks @edquist this looks solid

PR #121, opened 2018-04-26
Fix gfal dependencies; drop Gratia + osg-configure tests

brianhlin:
osg-configure unit tests don't really belong in the integration suite and we already run them in osg-configure's CI builds.

This fixes a fair amount of the 3.3 failures: http://vdt.cs.wisc.edu/tests/20180425-1629/results.html


matyasselmeci:
If it's not too hard, could you please separate this into multiple PRs? It would make things cleaner and we might change our minds on some of them.

brianhlin:
Mat gave me the go ahead to merge off-thread

PR #123, opened 2018-06-28
Python 3 compatibility for osg-test-log-viewer

matyasselmeci:
Tested with Python 2 and Python 3.4 on Fermicloud VMs, and Python 2.7 and 3.6 on my Fedora 28 laptop.

### DIFF
--- a/osg-test-log-viewer
+++ b/osg-test-log-viewer
@@ -38,7 +38,8 @@ def load_logdata_from_handle(handle):
 
     for line in handle:
         if six.PY3:
-            line = line.decode(errors="replace").rstrip()
+            line = line.decode(errors="replace")
+        line = line.rstrip()
         command_match = re.match(r'(?x)osgtest: \s* ' + datetime_re + r''': \s*
                                     (?P<test>[^:]+:[^:]+):
                                     (?P<lineno>\d+): \s*
--- a/osg-test-log-viewer
+++ b/osg-test-log-viewer
@@ -1,14 +1,16 @@
 #!/usr/bin/python
 """View the output from osg-test in a structured way"""
+from __future__ import print_function
 
-from Tkinter import *
-import ScrolledText
+import six
+from six.moves.tkinter import *
+from six.moves import tkinter_scrolledtext as ScrolledText
+from six.moves import tkinter_font as tkFont
+from six.moves import urllib
 import os
 import re
 import subprocess
 import sys
-import urllib2
-import tkFont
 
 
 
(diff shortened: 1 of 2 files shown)


osg-test-log-viewer
```diff
@@ -35,7 +37,8 @@ def load_logdata_from_handle(handle):
     current_test = ""
 
     for line in handle:
-        line = line.rstrip()
+        if six.PY3:
+            line = line.decode(errors="replace").rstrip()
```
edquist: I presume `six.PY3` checks whether python3 is running - but what happens if this is false?  (Does `.rstrip()` still get run if it's python2?)

matyasselmeci: Derp. Fixed.

edquist: Yay, the system works! :D

matyasselmeci:
We can probably drop Schrödinger's Python when we drop EL 7 support.

... in 2024.

bbockelm:
Python3.4 available within EPEL6, no?  At least theoretically, we could drop python2 support for minor tools since python3 is on all our platforms.

matyasselmeci:
I'd be okay with that; not sure how other people would feel.

There is some software we can't drop Python 2 support for yet because Python 3 versions of libraries they use aren't available on EL 6/EL 7 (I'm thinking osg-build and osg-configure, which use the RPM libraries).

bbockelm:
For something that's purely OSG S&R facing such as `osg-test-log-viewer` (or `osg-test` in general!), I think it's an experiment worth doing.  Worst case, we annoy other developers.

I agree that we can sit back and wait on dependency chains to improve in general.  I would suspect that RHEL7 would eventually get python3 versions of the RPM bindings.

That said, `osg-build` and `osg-configure` should likely do `six` work like this.

@brianhlin ?

edquist:
my quick thoughts - not really for or against -

- we can add a python3 requirement to osg-test to bring that in automatically (from epel if necessary)
- that might be annoying for unsuspecting users, but, for an internal tool, maybe that's not a problem
- installing osg-test from git does not automatically pull in the rpm requirements, including python3

brianhlin:
I'm defintely on board for sixifying our internal tools. I'm not sure `osg-configure` falls under this category.

> installing osg-test from git does not automatically pull in the rpm requirements, including python3

We need to update the `bootstrap-osg-test` to pull in all the requirements but it's very low on the `osg-test` priority list.

matyasselmeci:
BB is suggesting converting internal tools to Python 3 and sixifying external tools (the latter of which will probably be needed for EL 8 support).

PR #124, opened 2018-07-17
osg-test changes for SOFTWARE-3202

sthapa:
osg-test test changes in order to use slurm 17.  Last VMU results: 
http://vdt.cs.wisc.edu/tests/20180716-1750/results.html

### DIFF
--- a/osgtest/tests/test_29_slurm.py
+++ b/osgtest/tests/test_29_slurm.py
@@ -4,6 +4,8 @@ import osgtest.library.mysql as mysql
 import osgtest.library.osgunittest as osgunittest
 import osgtest.library.service as service
 
+import time
+
 CLUSTER_NAME = 'osg_test'
 CTLD_LOG = '/var/log/slurm/slurmctld.log'
 SHORT_HOSTNAME = core.get_hostname().split('.')[0]
@@ -39,6 +41,21 @@ SlurmdDebug=debug5
 StateSaveLocation=/var/spool/slurmd
 """
 
+SLURM_CGROUPS_CONFIG = """CgroupAutomount=yes
+CgroupMountpoint=/cgroup
+ConstrainCores=no
+ConstrainRAMSpace=no
+"""
+
+SLURM_CGROUPS_DEVICE_CONFIG = """/dev/null
+/dev/urandom
+/dev/zero
+/dev/sda*
+/dev/cpu/*/*
+/dev/pts/*
+"""
+
+
 class TestStartSlurm(osgunittest.OSGTestCase):
 
     def slurm_reqs(self):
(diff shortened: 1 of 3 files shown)


osgtest/tests/test_29_slurm.py
```diff
@@ -88,20 +119,27 @@ def test_03_start_slurm(self):
...
+            service.check_start(core.config['slurm.service-name'])
+            service.check_start(core.config['slurm.ctld-service-name'])
+        else:
+            service.check_start(core.config['slurm.service-name'])
 
         core.monitor_file(CTLD_LOG,
                           stat,
                           'slurm_rpc_node_registration complete for %s' % SHORT_HOSTNAME,
                           60.0)
+        time.sleep(10)
```
brianhlin: What exactly are we waiting for here? Can we use `core.monitor_file` to look for a sentinel instead?

osgtest/tests/test_29_slurm.py
```diff
@@ -52,6 +69,20 @@ def test_01_slurm_config(self):
                     SLURM_CONFIG % {'short_hostname': SHORT_HOSTNAME, 'cluster': CLUSTER_NAME, 'ctld_log': CTLD_LOG},
                     owner='slurm',
                     chmod=0644)
+        core.config['cgroup.config'] = '/etc/slurm/cgroup.conf'
+        config = SLURM_CGROUPS_CONFIG
+        if core.el_release() == 6:
+            config += "\nCgroupMountpoint=/cgroup"
+        files.write(core.config['cgroup.config'],
+                    SLURM_CGROUPS_CONFIG,
```
brianhlin: It looks like this should be `config` instead of `SLURM_CGROUPS_CONFIG`, otherwise the EL6 if block above doesn't really do anything

osgtest/tests/test_29_slurm.py
sthapa: When the slurm services are started, it looks like it takes them a little time to get fully operational.  The sleep it just to give the service a little time to get fully ready.  I'm not quite sure if or which bit of text gets emitted to indicate the service is ready.

osgtest/tests/test_29_slurm.py
sthapa: yeah, that's right, will update

osgtest/tests/test_29_slurm.py
brianhlin: Let's try and find that sentinel instead of adding a blanket 10s wait time

sthapa:
Test results with sleep replaced with a sentinel come up green: http://vdt.cs.wisc.edu/tests/20180724-1404/results.html

PR #125, opened 2018-07-25
allow installing slurm without all of osg-tested-internal (SOFTWARE-3…

edquist:
…347)


brianhlin:
One of the tests appear to have timed out when setting up the initial docker container. Merging.

PR #126, opened 2018-07-27
Fix htcondor-ce test_06 skip

sthapa:
Green on test results and test_06_slurm_trace runs with an ok instead of skipping

http://vdt.cs.wisc.edu/tests/20180727-1122/results.html

### DIFF
--- a/osgtest/tests/special_install.py
+++ b/osgtest/tests/special_install.py
@@ -33,8 +33,7 @@ class TestInstall(osgunittest.OSGTestCase):
         # FIXME: Install slurm out of contrib if we're running 'All' tests until
         # SOFTWARE-1733 gives us a generalized solution
         if 'osg-tested-internal' in pkg_repo_dict:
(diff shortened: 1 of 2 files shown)


osgtest/tests/special_install.py
```diff
@@ -33,8 +33,7 @@ def test_02_install_packages(self):
         # FIXME: Install slurm out of contrib if we're running 'All' tests until
         # SOFTWARE-1733 gives us a generalized solution
         if 'osg-tested-internal' in pkg_repo_dict:
-            all_slurm_packages = core.SLURM_PACKAGES + ['slurm-slurmdbd']
-            pkg_repo_dict.update(dict((x, ['osg-development']) for x in all_slurm_packages))
+            pkg_repo_dict.update(dict((x, ['osg-development']) for x in core.SLURM_PACKAGES))
```
brianhlin: Doesn't this mean that `slurm-slurmdbd` doesn't get installed?

PR #131, opened 2018-09-05
Update bootstrap-osg-test to fetch from source control (SOFTWARE-2232)

brianhlin:
I don't particularly care about the installing from RPM case


matyasselmeci:
Doesn't this not install rpm _dependencies_ of osg-test?

brianhlin:
osg-ca-generator is the only requirement for `osg-test`. This however, doesn't install the requirements for `osg-test-log-viewer` but I didn't think those were strictly necessary.

PR #132, opened 2018-09-05
set xrootd-lcmaps policy in sec_protocol (SOFTWARE-3396)

edquist:
rather than adding an xrootd_policy section to lcmaps.db

This would break due to an argument parsing bug in xrootd-lcmaps < 1.4.0

### DIFF
--- a/osgtest/tests/test_14_lcmaps.py
+++ b/osgtest/tests/test_14_lcmaps.py
@@ -20,16 +20,3 @@ class TestLcMaps(osgunittest.OSGTestCase):
                     "globus_mapping liblcas_lcmaps_gt4_mapping.so lcmaps_callout\n",
                     owner='lcmaps')
 
-    def test_02_xrootd_policy(self):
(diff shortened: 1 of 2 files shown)


edquist:
I'm going to kick off an osg-test run against this branch; i'll report the results here.

edquist:
VMU results are green:

http://vdt.cs.wisc.edu/tests/20180905-1751/packages.html

I also tried running the `xrdcp` command manually with xrootd-lcmaps 1.4.0 installed and it was successful, where it failed for 1.3.4. So anyway i think this should be good to go, unless you want to make this detect the xrootd-lcmaps version to still work for 1.3.4.

brianhlin:
> So anyway i think this should be good to go, unless you want to make this detect the xrootd-lcmaps version to still work for 1.3.4.

Nah, 1.4.0 is already released and fixes the issue so there's no reason to accommodate a bug in the tests.

PR #134, opened 2018-09-11
Fix direction of version comparison (SOFTWARE-3396)

edquist:
We only want to add this xrootd_policy section to lcmaps.db for
xrootd-lcmaps < 1.4.0, that is, skip if xrootd-lcmaps >= 1.4.0

Also, per Mat's suggestion, i am renaming the name of the test to
say that it is for the "old" xrootd_policy

...

Soon we'll add a more clear helper to do package version checks.


edquist:
I've also kicked off a VMU run for this

PR #138, opened 2018-11-19
Add another digit to tests so we can have more tests

matyasselmeci:


matyasselmeci:
Found it: https://github.com/opensciencegrid/vm-test-runs/blob/master/component-tags.yaml also needs to be updated.

PR #142, opened 2018-12-04
Shorten output by not displaying full path to file

matyasselmeci:
Also put the file name before the test name.  The resulting output will look like

`test_760_tomcat.py: test_01_stop_tomcat ... okskip`

as opposed to

`test_01_stop_tomcat (osgtest.tests.test_760_tomcat.TestStopTomcat) ... okskip`


matyasselmeci:
Sure, I can do that.

matyasselmeci:
Things are not showing up in the components display in the results. I'll have to fix that in vm-test-runs first (and make sure it works with both kinds of output).

matyasselmeci:
It does indeed break some parsing. It's not that important and I don't have time to fix it, so I'll close this PR.

PR #143, opened 2018-12-05
Fix test sequence (sorting was wrong)

matyasselmeci:


matyasselmeci:
We should rename the files, then.

brianhlin:
Sure that works now that you added more test numbers. Go ahead and do that in this PR

matyasselmeci:
Fixed, seems to work.

PR #144, opened 2018-12-06
Various fixups

matyasselmeci:


matyasselmeci:
Sure. 
`self.skip_ok_if(core.config['xrootd.gsi'] == "ON", 'fuse incompatible with GSI')`
always triggers because we set it in `test_01_start_xrootd` in `test_150_xrootd.py`.

brianhlin:
I have two concerns:

1. Should we be testing xrootd-fuse? We should reach out to ATLAS/CMS about usage
1. The incompatibility message is from 2012, is it still true?

matyasselmeci:
Closing this; we don't want the last commit and it looks like the other two got merged already.

PR #146, opened 2018-12-07
"exclude" gratia tests on el7+ instead of okskipping them

matyasselmeci:


edquist:
Looks fine, i think.  I forget though, do we eventually want to run these tests on el7 also?  I don't remember if we had disabled them because something was broken or because we planned never to support it.

matyasselmeci:
We couldn't get the server working on EL7 -- and I don't think we ever will, so unless we set up a mini-GRACC or something, those tests will remain forever disabled.

PR #147, opened 2018-12-07
StashCache tests (SOFTWARE-3360)

matyasselmeci:
Happy results: http://vdt.cs.wisc.edu/tests/20181207-1548/results.html


osgtest/tests/test_150_xrootd.py
```diff
@@ -58,7 +63,7 @@ def test_01_start_xrootd(self):
                             owner="xrootd",
                             chown=(user.pw_uid, user.pw_gid))
 
-            files.append(core.config['xrootd.config'], XROOTD_CFG_TEXT % sec_protocol, owner='xrootd', backup=True)
+            files.append(core.config['xrootd.config'], XROOTD_CFG_TEXT % (sec_protocol, core.config['xrootd.port']), owner='xrootd', backup=True)
```
brianhlin: Can we split this up so it fits within the 120 char limit?

matyasselmeci: Done

osgtest/tests/test_155_stashcache.py
```diff
@@ -0,0 +1,157 @@
...
+xrd.trace all
+cms.trace all
+""".format(**globals())
+
+
+CACHES_JSON_TEXT = """\
+[
+{"name":"root://localhost", "status":1, "longitude":-89.4012, "latitude":43.0731}
+]
+"""
```
brianhlin: Wow this is a lot of config. Can we simplify this after https://github.com/opensciencegrid/StashCache-Daemon/pull/21 gets released?

osgtest/tests/test_155_stashcache.py
```diff
@@ -0,0 +1,157 @@
...
+            ("origin_xroot_port", ORIGIN_XROOT_PORT),
+            ("cache_xroot_port", CACHE_XROOT_PORT)
+        ]:
+            _setcfg(key, val)
+
+        xrootd_user = pwd.getpwnam("xrootd")
+        for d in [_getcfg("origin_dir"), _getcfg("cache_dir"),
+                  os.path.dirname(_getcfg("caches_json_path"))]:
+            files.safe_makedirs(d)
+            os.chown(d, xrootd_user.pw_uid, xrootd_user.pw_gid)
```
brianhlin: What do you think about making `files.safe_makedirs` take optional args for chown/chmod (defaulting to `root:root` and `0644`)?

osgtest/tests/test_460_stashcache.py
```diff
@@ -0,0 +1,94 @@
...
+        ("testfile%d" % x, str(random.random()))
+        for x in range(4)
+    ]
+
+    def assertCached(self, name, contents):
+        fpath = os.path.join(_getcfg("cache_dir"), name)
+        self.assertTrue(os.path.exists(fpath),
+                        name + " not cached")
+        self.assertEqual(contents, files.read(fpath, as_single_string=True),
+                         "cached file %s contents do not match expected" % name)
```
brianhlin: Should we include the actual content vs the expected content in the message?

osgtest/tests/test_460_stashcache.py
```diff
@@ -0,0 +1,94 @@
...
+                        contents, backup=False, chmod=0o644,
+                        chown=(xrootd_user.pw_uid, xrootd_user.pw_gid))
+
+    def test_02_xroot_fetch_from_origin(self):
+        name, contents = self.testfiles[0]
+        result, _, _ = \
+            core.check_system(["xrdcp", "-d1", "-N", "-f",
+                               "root://localhost:%d//%s" % (_getcfg("origin_xroot_port"), name),
+                               "-"], "Checking xroot copy from origin")
+        self.assertEqual(result, contents, "downloaded file does not match expected")
```
brianhlin: Can we test HTTP fetches from the origin or is that not supported?

osgtest/tests/test_155_stashcache.py
matyasselmeci: Slightly. If we start with the cache server config from that PR (and it doesn't look like it has any origin server config changes), I'd need to make the following config changes as part of setup:

- set `cachedir` and `sitename` in `10-osg-site-local.cfg`
- change the HTTPS port in `40-osg-http.cfg`
- change the HTTP port in `40-osg-http.cfg`
- delete `40-osg-monitoring.cfg`
- change both `pss.origin` settings in `40-osg-xcache.cfg`

So basically we'd be looking at a bunch of `sed`-equivalents instead of one config file write. I suppose that's probably the right thing to do in the long run anyway...

osgtest/tests/test_155_stashcache.py
matyasselmeci: I'd rather have it follow the semantics of its namesake (`os.makedirs`) as opposed to making something different and unexpected.

But you just reminded me that the chown is redundant anyway since osg-test always runs as root.

osgtest/tests/test_460_stashcache.py
matyasselmeci: The origin isn't running the HTTP plugin.

osgtest/tests/test_460_stashcache.py
matyasselmeci: Sure. I added a new `assertEqualVerbose` method that prints the expected and actual values on failure.

osgtest/tests/test_155_stashcache.py
brianhlin: Maybe I'm misunderstanding your last comment but the chown that you removed is setting ownership to `xrootd:xrootd` and now it's chowning the dirs to `root:root`

osgtest/tests/test_155_stashcache.py
brianhlin: Would we be able to instead write a `99-local.cfg` that overrides the values in the files you mentioned or does XRootD config not work that way?

osgtest/tests/test_155_stashcache.py
matyasselmeci: /me looks at code again
I am a moron. Fixed.

osgtest/tests/test_155_stashcache.py
matyasselmeci: That's a good question. @bbockelm, if there are duplicate settings, do the later lines override the earlier ones?

matyasselmeci:
More happy tests: http://vdt.cs.wisc.edu/tests/20181211-0842/packages.html

osgtest/tests/test_155_stashcache.py
bbockelm: Depends on the configuration variable.

Xrootd configs are not a hash lookup (as with an INI file or HTCondor config) but rather linear parsing.  So, some attributes are additive (IIRC, the `all.export` lines are an example of this) where, with others, the last-specified wins.

I think everything you mention above is "last-specified wins".

Not clear - why does `40-osg-monitoring.cfg` need to be deleted though?  For the unit tests, I'd be really happy if we can stick to changing the configs in ways that we recommend to admins...

bbockelm: In that case, can you add a `-itb` suffix to the monitoring hostnames?  I don't think there's something listening on UDP there right now, but I'd like to have these show up eventually (as opposed to dropping them...)

matyasselmeci: Be careful what you wish for :P

So the new tasks will be:

- set `cachedir` and `sitename` in `10-osg-site-local.cfg`
- change the HTTPS port in `40-osg-http.cfg`
- change the HTTP port in `40-osg-http.cfg`
- replace xrd-*.osgstorage.org with xrd-*-itb.osgstorage.org in `40-osg-monitoring.cfg`
- change both `pss.origin` settings in `40-osg-xcache.cfg`

bbockelm: Only the first bullet is correct - for the others, we should be able to do this solely via a 99-overrides.cfg

bbockelm: Only the first bullet is correct - for the others, we should be able to do this solely via a 99-overrides.cfg

matyasselmeci: OK. Well, we'll do that for the next version of the tests.

PR #149, opened 2018-12-14
SOFTWARE-3359 Add Xrootd multiuser tests

efajardo:


osgtest/library/core.py
```diff
@@ -634,19 +634,25 @@ def get_hostname():
     except socket.error:
         return None
 
+def check_file_ownership(file_path, owner_name):
+    """Return True if at 'file_path' exists, is owned by
+    'owner_name' and is a  is a file
```
matyasselmeci: typo
```suggestion
    'owner_name' and is a file
```

osgtest/tests/test_450_xrootd.py
```diff
@@ -3,6 +3,7 @@
 import shutil
 import tempfile
 import pwd
+from os import stat
```
matyasselmeci: I don't think this is used anywhere since you're using `core.check_file_ownership` instead
```suggestion
```

osgtest/tests/test_450_xrootd.py
```diff
@@ -38,12 +38,20 @@ def test_01_xrdcp_local_to_server(self):
         fail = core.diagnose('xrdcp copy, local to URL',
                              command, status, stdout, stderr)
         file_copied = os.path.exists(os.path.join(temp_dir, 'copied_file.txt'))
-        shutil.rmtree(temp_dir)
-
+        if core.config['xrootd.multiuser'] != "ON":
+            shutil.rmtree(temp_dir)
```
matyasselmeci: What removes the temp dir if we _are_ doing multiuser?

efajardo:
@matyasselmeci thanks for reviewing can you please review again?

PR #153, opened 2018-12-31
PackageVersion comparison helper class

edquist:
This simplifies all of our package version checks with a helper class, which will compare the installed rpm version of a package to an `[E:]V[-R]` version string, in similar style with how version requirements are specified in rpm spec files.

The docstring for the new `PackageVersion` class provides usage examples.

This is a drop-in replacement for our (significantly less readable) existing version checks.

VMU test shows it looks fine, with the same results as the nightlies anyway:

http://vdt.cs.wisc.edu/tests/20181230-1900/packages.html

(... had this code sitting around since Sept, would like to finally get it in...)

### DIFF
--- a/osgtest/tests/test_410_jobs.py
+++ b/osgtest/tests/test_410_jobs.py
@@ -3,7 +3,6 @@
 #pylint: disable=R0904
 
 import os
-import rpm
 import shutil
 import tempfile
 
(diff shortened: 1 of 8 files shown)


osgtest/tests/test_410_jobs.py
```diff
@@ -49,11 +48,7 @@ def test_01_condor_run_pbs(self):
         # Figure out whether the installed BLAHP package is the same as or later
         # than "blahp-1.18.11.bosco-4.osg*" (in the RPM sense), because it's the
         # first build in which the job environments are correctly passed to PBS.
-        # The release following "osg" does not matter and it is easier to ignore
-        # the OS major version.  This code may be a useful starting point for a
-        # more general library function.
-        blahp_envra = core.get_package_envra('blahp')
-        blahp_pbs_has_env_vars = (rpm.labelCompare(['blahp', '1.18.11.bosco', '4.osg'], blahp_envra[1:4]) <= 0)
+        blahp_pbs_has_env_vars = core.PackageVersion('blahp') >= '1.18.11.bosco-4.osg'
```
brianhlin: We can actually drop this comparison since the earliest version of the blahp that we ship is 1.18.29. That can go in a different PR, though.

PR #158, opened 2019-01-23
Bump version to 2.3.1 (SOFTWARE-3532)

edquist:


brianhlin:
I'd like to hold off on this until we're closer to a release in case there are other osg-test changes required

PR #162, opened 2019-03-12
Vendor in rpmUtils.miscutils.stringToVersion() (it will probably be gone in RHEL 8)

matyasselmeci:
https://bugzilla.redhat.com/show_bug.cgi?id=1364504


edquist:
> it will probably be gone in RHEL 8

> probably

:D

PR #163, opened 2019-03-15
Rewrite stashcache tests to work with xcache >= 1.0.2

matyasselmeci:
Auth cache/origin tests don't work yet; also, to simplify the code, I dropped support for previous versions of xcache.

Tested this out on Fermicloud; need to do a VMU run too.

### DIFF
--- a/osgtest/tests/test_150_xrootd.py
+++ b/osgtest/tests/test_150_xrootd.py
@@ -31,6 +31,9 @@ u xrootd /tmp a
 
 class TestStartXrootd(osgunittest.OSGTestCase):
 
+    def setUp(self):
+        self.skip_ok_if(core.PackageVersion("xcache") >= "1.0.2", "xcache conflicts with xrootd tests")
+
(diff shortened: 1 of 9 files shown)


matyasselmeci:
Also, don't run this at the same time as the xrootd tests; they both use the same `/etc/xrootd/config.d` dir and would conflict hilariously with each other.

matyasselmeci:
Successful VMU run: http://vdt.cs.wisc.edu/tests/20190315-1436/packages.html

matyasselmeci:
New test run. http://vdt.cs.wisc.edu/tests/20190326-1706/packages.html

brianhlin:
Just so that it's noted somewhere, we plan to merge this after xcache is released

matyasselmeci:
In that case, can I add something to the VMU tests to use this branch to test the new stashcache?

matyasselmeci:
Never mind, we already have an xcache-1.0.2 branch.

PR #165, opened 2019-04-12
Get rid of old tests (SOFTWARE-3504)

matyasselmeci:
This gets rid of the 3.3-only tests from osg-test. Run here: http://vdt.cs.wisc.edu/tests/20190411-1609/packages.html; aside from random failures, the results are the same as the nightlies.

Once this goes in, I will tag the merge commit so we can resurrect or look at the old tests if we need to.


matyasselmeci:
I'm a bit confused. These tests are just the nightlies (e.g. http://vdt.cs.wisc.edu/tests/20190412-0423/packages.html) with osg-test from the new branch. Those do install stuff like bestman and gums but the new version doesn't test them.

edquist:
Are we getting rid of the osg-3.3 tests now in the nightlies?

matyasselmeci:
Not yet. We're getting rid of tests for components that are _only_ on 3.3.

osgtest/tests/test_580_gfal2util.py
```diff
@@ -20,60 +20,16 @@ def setUp(self):
     def get_srm_url_base(self):
         return 'srm://%s:%s/%s?SFN=' % (TestGFAL2Util.__hostname, TestGFAL2Util.__port, TestGFAL2Util.__sfn)
```
brianhlin: You should be able to get rid of this function, too

matyasselmeci:
I'll merge this tomorrow unless someone else has any objections. Edgar, are you OK with my explanation or do you want me to do something different?

matyasselmeci:
I saved the last commit before this was merged (2e7a903) as the tag "pre-osg33-purge".

PR #167, opened 2019-05-01
Don't run xrootd-tpc tests on EL6 and also factor out common 'xrootd' dependency

matyasselmeci:
The setUp method gets used for all the other test functions so common excludes/skip_oks can be placed there instead of being duplicated.

### DIFF
--- a/osgtest/tests/test_838_xrootd_tpc.py
+++ b/osgtest/tests/test_838_xrootd_tpc.py
@@ -5,12 +5,16 @@ import osgtest.library.osgunittest as osgunittest
 
 class TestStopXrootdTPC(osgunittest.OSGTestCase):
     @core.elrelease(7,8)
+    def setUp(self):
+        core.skip_ok_unless_installed("xrootd",
+                                      by_dependency=True)
+
     def test_01_stop_xrootd(self):
         if core.state['xrootd.tpc.backups-exist']:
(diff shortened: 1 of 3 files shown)


efajardo:
@matyasselmeci if Travis tests pass feel free to merge LGTM

efajardo:
@matyasselmeci I actually forgot to fix this. But non of the tpc tests should depend on: xrootd-scitokens
that was jsut before it was merged into xrootd package. Can you remove it from stop_tpc?

osgtest/tests/test_838_xrootd_tpc.py
```diff
@@ -5,12 +5,16 @@
...
+        core.skip_ok_unless_installed("xrootd",
+                                      by_dependency=True)
+
     def test_01_stop_xrootd(self):
         if core.state['xrootd.tpc.backups-exist']:
             files.restore(core.config['xrootd.tpc.config-1'], "xrootd")
             files.restore(core.config['xrootd.tpc.config-2'], "xrootd")
 
-        core.skip_ok_unless_installed('xrootd', 'xrootd-scitokens', by_dependency=True)
+        core.skip_ok_unless_installed('xrootd-scitokens', by_dependency=True)
```
efajardo: This should only depend on xrootd but since you did it before the dependency on scitokens should be removed.

efajardo:
LGTM

PR #168, opened 2019-05-03
Update tests to work with XCache 1.0.2+

matyasselmeci:


matyasselmeci:
Just kicked off a VMU run.

matyasselmeci:
VMU run: http://vdt.cs.wisc.edu/tests/20190503-1232/packages.html

PR #170, opened 2019-08-05
Support XCache 1.1 (SOFTWARE-3561)

brianhlin:
http://vdt.cs.wisc.edu/tests/20190802-1624/results.html


matyasselmeci:
Can you kick the broken travis test?

PR #176, opened 2019-08-21
Use /etc for the Slurm sysconfdir (SOFTWARE-3783)

brianhlin:
We could instead build Slurm with './configure
--sysconfdir=/etc/slurm' (the default is /etc) but I'd like to
minimize the differences between the upstream packaging


brianhlin:
None of the CI tests Slurm so I'm going to just merge this

PR #181, opened 2019-09-09
Add bad skip messages for tests that need a proxy cert

brianhlin:


matyasselmeci:
I think you should check if the cert is valid instead of whether it has been created.

brianhlin:
Sure but we don't have to hold back adding useful skip info to the test for that

PR #182, opened 2019-09-09
Check for proxy validity, not creation

matyasselmeci:


osgtest/tests/test_400_proxy.py
```diff
@@ -33,7 +33,8 @@ def test_03_grid_proxy_init(self):
         core.state['proxy.created'] = True
 
     def test_04_grid_proxy_info(self):
+        core.state['proxy.valid'] = False
         core.skip_ok_unless_installed('globus-proxy-utils')
-        self.skip_bad_unless(core.state['proxy.created'], 'Proxy creation failed')
```
brianhlin: We should probably still badskip if we didn't manage to create the proxy

matyasselmeci: What if a valid proxy already exists?

brianhlin: If there is one, it looks like we get rid of it in `test_01_remove_proxy`. So much for `osg-test` not messing with the pre-exsting environment :stuck_out_tongue:

matyasselmeci: Fair enough

PR #189, opened 2020-07-14
HAve third party copy servers use their HTTP ports

efajardo:
HAve third party copy servers use their HTTP ports as Xrootd ports to prevent port collision


brianhlin:
@efajardo mat's out today. Can you kick off VMU tests based off this branch to see if this fixes the problems you were seeing?

PR #191, opened 2020-07-22
EL8/Python 3 support

matyasselmeci:


matyasselmeci:
Could you hold off on merging this? I need to check the voms change.

matyasselmeci:
Yeah it's wrong but doesn't matter: `-nameopt compat` mean the opposite things in openssl 1.0 and openssl 1.1 but the field that's getting parsed isn't affected one way or the other.

osgtest/tests/test_550_condorce.py
```diff
@@ -126,9 +129,9 @@ def test_07_ceview(self):
         core.skip_ok_unless_installed('htcondor-ce-view')
         view_url = 'http://%s:%s' % (core.get_hostname(), int(core.config['condor-ce.view-port']))
         try:
-            src = urllib2.urlopen(view_url).read()
-        except urllib2.URLError:
-            self.fail('Could not reach HTCondor-CE View at %s' % view_url)
+            src = to_str(urlopen(view_url).read())
+        except EnvironmentError as err:
+            self.fail('Could not reach HTCondor-CE View at %s: %s' % (view_url, err))
```
brianhlin, line 134: Huh, this shows up as `EnvironmentError` now?

osgtest/tests/test_280_gsiopenssh.py
```diff
@@ -34,9 +34,9 @@ def test_01_set_config(self):
         core.state['gsisshd.can-run'] = (not (
             core.el_release() >= 7 and
             core.state['selinux.mode'] and
-            not core.rpm_is_installed('policycoreutils-python')))
+            not core.dependency_is_installed("/usr/sbin/sestatus")))
         self.skip_ok_unless(core.state['gsisshd.can-run'],
-                            "Can't run with SELinux on EL >= 7 without policycoreutils-python")
+                            "Can't run with SELinux on EL >= 7 without %s" % TestStartGSIOpenSSH.policycoreutils_rpm)
```
brianhlin, line 39: This doesn't appear to be defined anywhere

osgtest/tests/test_550_condorce.py
matyasselmeci: In Python 2 it's `urllib2.URLError` and in Python 3 it's `urllib.error.URLError`. Both are subclasses of EnvironmentError (which is also the parent of IOError and OSError, so it's a good catch-all for ... basically errors caused by something outside of your program). I could add URLError to the imports if you want me to be more specific.

brianhlin: Nope it's fine, I was just curious more than anything!

matyasselmeci: I divide exceptions into two categories: "my fault" and "not my fault" :-)

PR #195, opened 2020-08-10
Wait to make sure xrootd starts up and dump log file in case of failure

matyasselmeci:


matyasselmeci:
The stashcache test set failed, but that was expected: https://travis-ci.com/github/opensciencegrid/osg-test/jobs/370485486. I'm going to merge.

PR #196, opened 2020-08-14
Use set to avoid duplicate transaction ids on upgrade tests too

matyasselmeci:


matyasselmeci:
The stashcache test failure is a separate thing I've been working on.

matyasselmeci:
VMU: https://osg-sw-submit.chtc.wisc.edu/tests/20200814-1200/packages.html
The "0 0 0" test seems like it's misreported -- looks OK from the logs.

PR #198, opened 2020-08-19
Missed a few str/bytes conversions

matyasselmeci:


matyasselmeci:
Exactly.
```
stashcp'ed file mismatch (actual b'0.12999031502581582\n' != expected '0.12999031502581582\n')
```

PR #202, opened 2020-10-05
Also install x509-scitokens-issuer-client out of development if we're…

matyasselmeci:
… doing osg-tested-internal


brianhlin (commented):
@efajardo Should we install out of contrib?

matyasselmeci:
Contrib is older than development -- 0.5.3 vs 0.8.0.

brianhlin (approved):
LGTM

brianhlin:
Re-kicking the failed 3.5 StashCache

PR #205, opened 2020-10-28
HTCondor starts up too fast in 8.9.9 and we miss the schedd startup

brianhlin:
`core.get_stat` returns `None` if the file doesn't exist and `condor.wait_for_daemon` handles `None` stats by looking for the relevant sentinel starting at the beginning of the file. Things look pretty good here: https://osg-sw-submit.chtc.wisc.edu/tests/20201028-0817/results.html


matyasselmeci:
How does this behave with multiple osg-test runs, when htcondor-ce has already been started and stopped once?

brianhlin:
1. If the CE service was started outside of osg-test, we don't end up using the stat. I could move the stat after this check so we don't bother with it in that case
1. If the CE service is not running, we grab the current stat and check for the schedd reporting back to the collector starting from the stat

matyasselmeci (approved):

PR #208, opened 2020-10-29
Add contents of CEViewLog to test output

brianhlin:


osgtest/tests/test_550_condorce.py
```diff
@@ -131,7 +138,13 @@ def test_07_ceview(self):
         try:
             src = core.to_str(urlopen(view_url).read())
         except EnvironmentError as err:
-            self.fail('Could not reach HTCondor-CE View at %s: %s' % (view_url, err))
+            debug_contents = 'Could not reach HTCondor-CE View at %s: %s\n%s\n' % (view_url, err, '=' * 20)
+            debug_file = '/var/log/condor-ce/CEViewLog'
+            try:
+                debug_contents += files.read(debug_file, True)
+            except FileNotFoundError:
```
matyasselmeci, line 145:
```suggestion
            except EnvironmentError:
```
EnvironmentError is still present in Python 3, and FileNotFoundError is a subclass of it, so this will work on both.

Note that on Python 2, a file not found is an OSError, not an IOError. EnvironmentError is a parent of both.

osgtest/tests/test_550_condorce.py
```diff
@@ -131,7 +138,13 @@ def test_07_ceview(self):
...
+                debug_contents += 'Failed to read %s\n' % debug_file
+            self.fail(debug_contents)
```
matyasselmeci, line 147: I'm not a fan of a log dump in a failure message. Failure messages should be limited to one line.

Just cat the file so it appears in its own section.

brianhlin: > Just cat the file so it appears in its own section.

Like in a separate test or just in a separate command within the same test?

matyasselmeci: I think a separate command is fine.

brianhlin: Done

matyasselmeci (approved):
Well, we'll see how it looks. Feel free to merge once the CI passes.

brianhlin:
A singularity test got wedged but other than that things look good. Merging

PR #214, opened 2021-03-25
Update Slurm tests for Slurm 20 (SOFTWARE-4446)

brianhlin:
Test results here: https://osg-sw-submit.chtc.wisc.edu/tests/20210325-1148/packages.html. 3.6 `slurm_trace` tests fail because the 3.6 blahp is broken but this otherwise fixes all the other Slurm startup failures we're seeing in the nightlies


brianhlin:
What's the worst that could happen? Merging

PR #221, opened 2021-05-05
Use new config for newer versions of xrootd-multiuser

matyasselmeci:
(SOFTWARE-4599)

Example tests are part of https://osg-sw-submit.chtc.wisc.edu/tests/20210505-1714/packages.html (see the "XRootD" package set).

### DIFF
--- a/osgtest/tests/test_150_xrootd.py
+++ b/osgtest/tests/test_150_xrootd.py
@@ -110,7 +110,11 @@ class TestStartXrootd(osgunittest.OSGTestCase):
 
     def test_03_configure_multiuser(self):
         core.skip_ok_unless_installed('xrootd-multiuser', 'globus-proxy-utils', by_dependency=True)
-        xrootd_multiuser_conf = "xrootd.fslib libXrdMultiuser.so default"
+        if core.PackageVersion("xrootd-multiuser") < "1.0.0-0":
+            xrootd_multiuser_conf = "xrootd.fslib libXrdMultiuser.so default"
+        else:
+            xrootd_multiuser_conf = "ofs.osslib ++ libXrdMultiuser.so\n" \
+                                    "ofs.ckslib ++ libXrdMultiuser.so"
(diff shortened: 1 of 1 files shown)


matyasselmeci:
These are config lines that are provided as part of the xrootd-multiuser RPM itself so in theory we don't have to change any documentation for this.

brianhlin:
> These are config lines that are provided as part of the xrootd-multiuser RPM itself so in theory we don't have to change any documentation for this.

Ultimately I'd prefer relying on the default config provided by the packaging wherever we can. Are we not using `/etc/xrootd/config.d/` in the tests?

matyasselmeci:
Nope.  All our xrootd configs are hand-written, single files.

One of the problems with config.d is that we install multiple xrootd packages but test the features individually, but there's only one config.d, which is shared across all of them.

To use stashcache as an example, we need both the cache and the origin RPMs installed to be able to test, but both RPMs just place files in config.d. If we tried to use that, we'd end up starting some sort of misshapen cacheorigin chimera that would totally not work.

What would have been nice (again for my stashcache example) is if we had created separate config.d dirs for the different services (e.g. `/etc/xrootd/stash-cache-config.d` and `stash-origin-config.d`) but it's probably too late to do that (for OSG 3.6 anyway). We could perhaps compose our test configs by cherry-picking individual files from config.d but that's brittle to renames and it would be really hard to get an overview of exactly what config we're testing...

brianhlin:
Good point. Maybe we can implement the per-service config.d just in the tests, then? It should be easy enough to take the entrypoint config (e.g., `/etc/xrootd/xrootd-stash-cache.cfg`), mangle the include dir, and copy all the relevant files from the package (i.e., `rpm -ql <package> | grep config.d`) to the service-specific config.d

brianhlin (approved):
LGTM for now. Do you have working VMU results with this change?

matyasselmeci:
We could try that.
Test results are in my first comment:
> Example tests are part of https://osg-sw-submit.chtc.wisc.edu/tests/20210505-1714/packages.html (see the "XRootD" package set).

PR #224, opened 2021-05-20
Revert "The testing should include multistream"

matyasselmeci:
This has been broken with XRootD 5 for a long time, but multistream TPC is not a common OSG use case

This reverts commit c4d10e044fd1a56a144cb2085c7b19baef827cf9.


brianhlin:
It's been so long that it feels wrong to see green XRootD tests

brianhlin (approved):
LGTM!

PR #227, opened 2021-06-16
Fix KeyError in TPC tests if macaroon creation failed (SOFTWARE-4604)

matyasselmeci:
Note: this doesn't solve any test failures, just gives us a better error message.
(SOFTWARE-4604)


brianhlin (approved):
LGTM!

PR #228, opened 2021-08-11
Turn core.state into a defaultdict with values deafulting to False

brianhlin:
The only state keys that aren't bools are the ones used for https://github.com/opensciencegrid/osg-test/blob/master/osgtest/tests/special_install.py


matyasselmeci:
Much as I hate KeyErrors, using a regular dict as core.state makes typos easier to catch.

brianhlin:
Good point

PR #229, opened 2021-08-12
Use BEARER_TOKEN_FILE env var to specify the token path

brianhlin:
Token discovery details: https://zenodo.org/record/3937438. Support was added for this to `condor_submit` in 9.0.2 and is expected in 9.1.3. I'm working on making use of this in `condor_ce_run` and adding token discovery to `condor_ce_trace`

### DIFF
--- a/osgtest/tests/test_410_jobs.py
+++ b/osgtest/tests/test_410_jobs.py
@@ -12,6 +12,16 @@ import osgtest.library.osgunittest as osgunittest
 
 class TestRunJobs(osgunittest.OSGTestCase):
 
+    def tearDown(self):
+        env_vars = ('_condor_SCITOKENS_FILE',
+                    'BEARER_TOKEN_FILE')
+
+        for env_var in env_vars:
+            try:
+                os.environ.pop(env_var)
+            except KeyError:
+                pass
+
     def verify_job_environment(self, output):
         expected_env = {'JOB_ENV': 'vdt',
                         'LOCAL_JOB_ENV': 'osg'}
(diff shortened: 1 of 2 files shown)


osgtest/tests/test_410_jobs.py
```diff
@@ -12,6 +12,16 @@
 
 class TestRunJobs(osgunittest.OSGTestCase):
 
+    def tearDown(self):
+        env_vars = ('_condor_SCITOKENS_FILE',
+                    'BEARER_TOKEN_FILE')
+
+        for env_var in env_vars:
+            try:
+                os.environ.pop(env_var)
```
matyasselmeci, line 21:
```suggestion
                del os.environ[env_var]
```
pop() is only if you also want the value back.
Alternatively you could use
```
os.environ.pop(env_var, None)
```
in which case you can drop the KeyError check

brianhlin: Fixed! The fewer tabs the better

osgtest/tests/test_410_jobs.py
```diff
@@ -12,6 +12,13 @@
 
 class TestRunJobs(osgunittest.OSGTestCase):
 
+    def tearDown(self):
+        env_vars = ('_condor_SCITOKENS_FILE',
+                    'BEARER_TOKEN_FILE')
+
+        for env_var in env_vars:
+            os.environ.pop(env_var, None)
```
edquist, line 20: I was going to make the same comment about `del d[k]`.  Looks great how you've done it though.

> The fewer tabs the better

well if that's the case...

```suggestion
        [ os.environ.pop(env_var, None) for env_var in env_vars ]
```

(whose returned list will also be thrown to the wind)

edquist (approved):
lgtm

PR #232, opened 2021-08-29
Use https for the osg-ca-generator submodule so I can pull it in on machines where I don't have my private key

matyasselmeci:


brianhlin (approved):
Sure, LGTM

PR #233, opened 2021-09-08
Don't print output for some system calls

matyasselmeci:
Compare https://osg-sw-submit.chtc.wisc.edu/tests/20210903-2358/results.html
to a nightly run https://osg-sw-submit.chtc.wisc.edu/tests/20210904-0423/results.html

doesn't look much different in the raw logs but if you're doing a manual run with --df (to dump output into a separate file) or are looking at the output with osg-test-log-viewer, it's nice not to have it cluttered up with a ton of "rpm --query" and "systemctl is-active" calls.

### DIFF
--- a/osgtest/library/core.py
+++ b/osgtest/library/core.py
@@ -270,7 +270,7 @@ def monitor_file(filename, old_stat, sentinel, timeout):
     return (None, None)
 
 
-def system(command, user=None, stdin=None, log_output=True, shell=False, timeout=None, timeout_signal='TERM'):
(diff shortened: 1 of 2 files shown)


brianhlin:
We still print the output on their failure, right?

matyasselmeci:
The two things being excluded are `systemctl is-active` and `rpm --query`, where we do not distinguish between failure (whatever that can be) and a negative result (thing is not running/package is not installed). So no.

brianhlin (approved):
Oh in that case, full steam ahead

PR #235, opened 2021-09-22
Account for new gfal2-util package split (SOFTWARE-4825)

brianhlin:
Test results here https://osg-sw-submit.chtc.wisc.edu/tests/20210922-0855/packages.html. EL8 failures are expected because only `python3-gfal2-util` is available on EL8 and it doesn't provide `gfal2-util`

### DIFF
--- a/osgtest/tests/test_580_gfal2util.py
+++ b/osgtest/tests/test_580_gfal2util.py
@@ -13,7 +13,8 @@ class TestGFAL2Util(osgunittest.OSGTestCase):
 
     def setUp(self):
         self.skip_ok_unless(core.state['proxy.valid'] or core.state['voms.got-proxy'], "No proxy")
-        core.skip_ok_unless_installed('gfal2-util', 'gfal2-plugin-file')
+        core.skip_ok_unless_installed('gfal2-util-scripts', 'gfal2-plugin-file')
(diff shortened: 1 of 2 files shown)


osgtest/tests/test_580_gfal2util.py
```diff
@@ -13,7 +13,8 @@ class TestGFAL2Util(osgunittest.OSGTestCase):
 
     def setUp(self):
         self.skip_ok_unless(core.state['proxy.valid'] or core.state['voms.got-proxy'], "No proxy")
-        core.skip_ok_unless_installed('gfal2-util', 'gfal2-plugin-file')
+        core.skip_ok_unless_installed('gfal2-util-scripts', 'gfal2-plugin-file')
+        core.skip_ok_unless_one_installed('python2-gfal2-util', 'python3-gfal2-util')
```
edquist, line 17: So what you apparently figured out, but was not obvious to me from the commit message or PR body, is that the new `gfal2-util-scripts` package contains the (non-python) shell scripts from `gfal2-util`, and that the new `python[23]-gfal2-util` packages now pull in `gfal2-util-scripts`.

But yeah with that in mind the updated logic looks good!

edquist (approved):

osgtest/tests/test_580_gfal2util.py
brianhlin: I'll update the commit message to clarify!

PR #236, opened 2021-09-24
Add "force" argument to service functions to ignore core.state flags and always start/stop the service

matyasselmeci:


brianhlin (approved):
LGTM

PR #237, opened 2021-09-24
Move skip_bad_unless_running() out of test_460_stashcache.py since it's more generally useful

matyasselmeci:
This does a BadSkip if one of the listed services is not running.

### DIFF
--- a/osgtest/library/condor.py
+++ b/osgtest/library/condor.py
@@ -1,6 +1,6 @@
 import os
 
-import osgtest.library.core as core
+from osgtest.library import core
 
 def wait_for_daemon(collector_log_path, stat, daemon, timeout):
     """Wait until the requested 'daemon' is available and accepting commands by
--- a/osgtest/library/files.py
+++ b/osgtest/library/files.py
@@ -26,7 +26,7 @@ import re
 import shutil
 import tempfile
 import hashlib
-import osgtest.library.core as core
+from osgtest.library import core
 
 
 _backup_directory = '/usr/share/osg-test/backups'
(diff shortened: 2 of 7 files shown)


matyasselmeci:
Green results: https://osg-sw-submit.chtc.wisc.edu/tests/20210923-2258/packages.html

Without the import changes, I got results like https://osg-sw-submit.chtc.wisc.edu/tests/20210923-2213/packages.html.

brianhlin (approved):
Huh, that import syntax change is weird (why didn't it show up before?). Did you intend to squash the last two commits?

Otherwise, LGTM. If you didn't want to squash the commits, go ahead and merge.

matyasselmeci:
Squashed.
I think the issue is that this commit introduces a circular import: `osgunittest` imports `service`, which imports `core`, which imports `osgunittest`, and the semantics of import resolution change somehow.

Note 1: now that we're using Python 3, we can use even nicer import syntax:
```
from ..library import core, files, service
```

Note 2: `core`'s dependency on `osgunittest` is because of the various `skip_*` functions in `core` (`skip_ok_unless_installed()` etc.). To resolve the loop, what I would actually like to do is move all of those functions into `OSGUnitTest` so we don't have to remember "do I use `core.` or `self.` for this skip function" all the time. That's a more extensive change though.

brianhlin:
```
from ..library import core, files, service
```

As annoying as I find Python 3 sometimes, this is one of my favorite things


> Note 2: core's dependency on osgunittest is because of the various skip_* functions in core (skip_ok_unless_installed() etc.). To resolve the loop, what I would actually like to do is move all of those functions into OSGUnitTest so we don't have to remember "do I use core. or self. for this skip function" all the time. That's a more extensive change though.

Agreed

PR #238, opened 2021-10-05
Get rid of the inner json.dumps() when constructing the payload for demo.scitokens.org

matyasselmeci:
https://osg-sw-submit.chtc.wisc.edu/tests/20211005-1035/packages.html


brianhlin (approved):
LGTM

PR #240, opened 2021-10-20
Refactor scitokens stuff into a separate library and request scitokens for xrootd

matyasselmeci:
Create a `library/credentials.py` library that contains functions for requesting and manipulating credentials (currently scitokens but in the future also idtokens and macaroons); modify `test_401_scitokens.py` to use this new library.

The primary function in credentials is `request_scitoken()` which requests a scitoken from demo.scitokens.org and sets values in the `core.state` and `core.config` dict appropriately.  Created tokens are cleaned up in `special_cleanup.py`.

`core.state['token.NAME_created']` now really means 'did we create this token' instead of 'does this token exist'; condor tests now use `os.path.exists()` instead.

### DIFF
--- a/osgtest/library/credentials.py
+++ b/osgtest/library/credentials.py
@@ -66,7 +66,11 @@ def parse_scitoken(token_string: str):
 
 
 def reserve_scitoken(token_name: str, token_file: str = None):
-    """Create the global dictionary entries used for storing/referencing scitokens"""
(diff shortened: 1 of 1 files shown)


matyasselmeci:
VMU run forthcoming...

brianhlin:
Yup, fine by me

PR #241, opened 2021-10-21
XRootD for 3.6 tests

matyasselmeci:
Fairly large overhaul of the xrootd test code; this creates a directory structure with which it can test both authenticated and unauthenticated access.

SciTokens are used for tests on 3.6; GSI/Macaroons are used for tests on 3.5. I also added a test to run `cconfig` and make sure that the overall xrootd config, which is assembled from multiple config files, contains the right values for some of the knobs we care about.

Known issues:
- FUSE tests are broken (unknown failure)
- Macaroons tests are broken (the destination token does not have the right permissions)
- SciTokens tests are only run on 3.6, not 3.5+upcoming -- this is because 3.5's osg-xrootd-standalone sets up GSI
- xrdvoms/vomsxrd/xrootd-voms-plugin is not tested yet

The commits in this PR are not in chronological order; instead I grouped changes into chunks and made commits for those in order to try and make a consistent "story"; especially the "Reorganize test directory to check privileges" series of commits should be taken as a whole (don't expect a revert to work). The first few commits (up to "Add logging config...") are probably independent.

VMU test forthcoming.

### DIFF
--- a/osgtest/library/files.py
+++ b/osgtest/library/files.py
@@ -195,10 +195,12 @@ def append(path, contents, force=False, owner=None, backup=True):
     write(path, new_contents, backup=False)
 
 
-def restore(path, owner):
+def restore(path, owner, ignore_missing=False):
     """Restores the path to its state prior to being written by its owner."""
     backup_id = (path, owner)
     if backup_id not in _backups:
+        if ignore_missing:
+            return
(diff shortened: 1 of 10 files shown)


matyasselmeci:
Also this depends on the tokens created in #240 (though it is not merged into this branch).

PR #243, opened 2021-11-01
Use voms-proxy-init to get grid proxies if needed

matyasselmeci:
Some tests (even in OSG 3.6) need grid proxies which we can get with
voms-proxy-init if grid-proxy-init is not available.

Also do some cleanup and defensive coding for VOMS.


brianhlin (approved):
LGTM. Got tests to back it up?

matyasselmeci:
Just manual stuff; this is cherry-picked from my xrootd branch.

PR #244, opened 2021-11-01
More context managers

matyasselmeci:
- Change environ_context to set multiple environment variables
- Add no_x509 context manager for temporarily moving the user's x509 credentials out of the way
- Add core.no_bearer_token() context manager for testing without a token


brianhlin (approved):
LGTM

PR #246, opened 2021-11-03
Use voms-proxy-init to get certs attempt 2

matyasselmeci:
Same objective as #243 except not broken. This time it checks if a user cert/key is available and doesn't try to make the proxy if it isn't (okskip).

I also have tests this time: https://osg-sw-submit.chtc.wisc.edu/tests/20211102-1152/packages.html
I replaced 'globus-proxy-utils' with 'voms-clients-cpp' in all the parameters files for this test. TPC failed to start because the config file was only created if globus-proxy-utils is installed -- obviously that only happens with this param set and is fixed in my branch anyway.

### DIFF
--- a/osgtest/tests/special_user.py
+++ b/osgtest/tests/special_user.py
@@ -102,6 +102,7 @@ class TestUser(osgunittest.OSGTestCase):
         # Set up certificate
         globus_dir = os.path.join(core.state['user.pwd'].pw_dir, '.globus')
         core.state['user.cert_path'] = os.path.join(globus_dir, 'usercert.pem')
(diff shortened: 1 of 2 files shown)


brianhlin:
> TPC failed to start because the config file was only created if globus-proxy-utils is installed -- obviously that only happens with this param set and is fixed in my branch anyway.

Fixed in which branch? It'd be nice to include that fix in this PR if possible

matyasselmeci:
Fixed in the big xrootd-for-3.6 branch.

brianhlin (approved):
Ah I understand your point about the param set now. LGTM

PR #249, opened 2022-02-23
Add some more fetch-crl warnings to the ignore list

matyasselmeci:
Alternatively we could just ignore `^VERBOSE\(0\)` since it's informational.


brianhlin:
> Alternatively we could just ignore ^VERBOSE\(0\) since it's informational.

Yeah, let's do it

brianhlin (approved):
LGTM

PR #251, opened 2022-03-11
Enable voms if the appropriate osg-xrootd version is installed

matyasselmeci:
https://osg-sw-submit.chtc.wisc.edu/tests/20220310-1741/results.html


brianhlin (approved):
Great variable name 😄

PR #252, opened 2022-04-14
allow blahp to be installed by dependency (SOFTWARE-4964)

edquist:


timtheisen (approved):
I think that this looks good.

PR #254, opened 2022-10-21
Compare and print source/destination file sizes for XRootD-TPC tests

matyasselmeci:
since it would be useful if a mismatch means the output is garbage, truncated, or empty.


edquist (approved):
LGTM.

PR #259, opened 2023-03-28
Use default cgroup location, /sys/fs/cgroup (SOFTWARE-5500)

brianhlin:
The old config resulted in errors like the following in slurm.log:

```
[2023-03-27T04:49:43.934] error: cannot read
/cgroup/system.slice/slurmd.service/cgroup.controllers: No such file or
directory
```

Success here https://osg-sw-submit.chtc.wisc.edu/tests/20230328-1359/results.html


matyasselmeci:
wow why did that ever work

matyasselmeci (approved):

PR #260, opened 2023-06-14
Serve a local namespaces json file for testing stashcp with our own namespaces

matyasselmeci:
stashcp 1.12.1+ bails out if you try to download from an unregistered namespace (according to a namespaces json endpoint of Topology). We can't tell stashcp to use a local namespaces.json file, but we can tell it to use a different Topology server. This PR runs a tiny HTTP server that serves the namespace info that works with our tests.


osgtest/tests/test_835_stashcache.py
```diff
@@ -50,3 +56,17 @@ def test_07_remove_certs(self):
...
+            proc.terminate()
+            try:
+                proc.wait(5)
+            except subprocess.TimeoutExpired:
+                proc.kill()
+        try:
+            logfile = getcfg("namespaces_json_server_logfile")
+            os.unlink(logfile)
+        except (KeyError, OSError):
+            pass
```
brianhlin, line 72: Missing newline

osgtest/tests/test_155_stashcache.py
```diff
@@ -213,3 +220,24 @@ def test_04_start_stash_cache(self):
...
+
+        # Make sure it didn't immediately crash
+        time.sleep(2)
+        ret = proc.poll()
+        if ret is not None:
+            core.system(["/bin/cat", logfile])
+            self.assertEqual(ret, None, f"namespaces JSON server terminated prematurely with code {ret}")
+
+        setstate("namespaces_json_server_proc", proc)
+        setcfg("STASH_NAMESPACE_URL", "http://localhost:1080")
```
brianhlin, line 243: I can see this test being generally useful in the future. It's a good candidate for breaking out into one of the `osg-test` libs!

brianhlin (approved):
Very minor comments. Pre-approving

osgtest/tests/test_155_stashcache.py
matyasselmeci: I'll break it out if I use it again; I probably could have gotten rid of most of this if I had just made a systemd .service file for namespaces_json_server

brianhlin: > I'll break it out if I use it again

Yeah, that's what I was thinking

> I probably could have gotten rid of most of this if I had just made a systemd .service file for namespaces_json_server

I did have the same thought here. It's probably less work to just create the systemd .service, isn't it? If you agree, let's just go with this implementation now and add a `# TODO` comment to turn this into a systemd service later

matyasselmeci: OK, TODOs added.

PR #263, opened 2023-09-11
Fix cleanup failures when using osg-update-data (since it depends on osg-release)

matyasselmeci:
(orig. title was "Do not erase osg-release before upgrading")
It is not necessary and breaks packages that depend on osg-release like osg-update-vos.

### DIFF
--- a/osgtest/library/files.py
+++ b/osgtest/library/files.py
@@ -219,6 +219,7 @@ def restore(path, owner, ignore_missing=False):
 
 def remove(path, force=False):
     """Remove the path, which could be a file, symlink, empty directory, or file glob.
+    Does nothing if the path is already missing.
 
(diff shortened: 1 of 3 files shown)


matyasselmeci:
Let me do a test run first...

matyasselmeci:
OK, this PR now fixes the cleanup failures by passing the flag to rpm to ignore dependencies when erasing the package.  There was a separate cleanup error caused by trying to remove the OSG-Test CA certs when /etc/grid-security/certificates had already been removed (by uninstalling a package that owned the directory).

matyasselmeci:
Test results: https://osg-sw-submit.chtc.wisc.edu/tests/20230913-1326/packages.html
This is ready for review.

brianhlin (approved):
LGTM

PR #264, opened 2023-09-14
Prepare for OSG 23 (SOFTWARE-5679)

brianhlin:
grep for '3.' didn't really turn up much of interest

### DIFF
--- a/osgtest/tests/test_380_cvmfs.py
+++ b/osgtest/tests/test_380_cvmfs.py
@@ -11,7 +11,7 @@ import osgtest.library.osgunittest as osgunittest
 class TestCvmfs(osgunittest.OSGTestCase):
 
     __check_path = '/cvmfs/cms.cern.ch/cmsset_default.sh'
-    __cvmfs_image = '/cvmfs/singularity.opensciencegrid.org/opensciencegrid/osg-wn:3.3-el6'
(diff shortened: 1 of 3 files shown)


osgtest/tests/test_380_cvmfs.py
```diff
@@ -11,7 +11,7 @@
 class TestCvmfs(osgunittest.OSGTestCase):
 
     __check_path = '/cvmfs/cms.cern.ch/cmsset_default.sh'
-    __cvmfs_image = '/cvmfs/singularity.opensciencegrid.org/opensciencegrid/osg-wn:3.3-el6'
+    __cvmfs_image = '/cvmfs/singularity.opensciencegrid.org/opensciencegrid/osgvo-el9:latest'
```
matyasselmeci, line 14: what's this image used for? osgvo-el9 is 778 MB, about 3 times the size of osg-wn:3.3-el6 (221 MB).  If all we need is a runnable image then perhaps we could use library/alpine:latest

matyasselmeci (approved):
one comment; pre-approving
EDIT: question, not comment

osgtest/tests/test_380_cvmfs.py
brianhlin: Yeah, that's a good idea

PR #265, opened 2023-09-22
mod_auth_oidc is now available in appstream on EL8/9 (SOFTWARE-5679)

brianhlin:
This was causing EL9 test failures for the central collector:

- Before: https://osg-sw-submit.chtc.wisc.edu/tests/20230921-1431/results.html
- After: https://osg-sw-submit.chtc.wisc.edu/tests/20230921-1736/results.html


matyasselmeci (approved):
Sweet

PR #267, opened 2023-11-10
Get stashcp tests working with Pelican

matyasselmeci:
- The `pelican-osdf-compat` RPM doesn't provide the `stashcache-client` virtual dependency; it's a very old name (used for the Python version of `stashcp`) so check for the `osdf-client` virtual dependency instead when deciding whether to skip the `stashcp` test or not

- Pelican uses `STASH_TOPOLOGY_NAMESPACE_URL` instead of `STASH_NAMESPACE_URL` for the URL to the namespaces JSON; set both in the environment so old and new versions of `stashcp` work as expected

- Better detect if we're using a Go version of `stashcp` (and should pass `--cache` instead of editing `caches.json`)

### DIFF
--- a/osgtest/tests/test_155_stashcache.py
+++ b/osgtest/tests/test_155_stashcache.py
@@ -145,7 +145,7 @@ class TestStartStashCache(OSGTestCase):
     def setUp(self):
         core.skip_ok_unless_installed("stash-origin",
                                       "stash-cache",
-                                      "stashcache-client",
+                                      "osdf-client",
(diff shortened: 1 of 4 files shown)


matyasselmeci:
Tests here: https://osg-sw-submit.chtc.wisc.edu/tests/20231110-1829/packages.html
The "23 Testing + Upcoming Testing" tests used Pelican stashcp.

timtheisen (approved):
LGTM

PR #268, opened 2024-02-07
Require "stashcp" for stashcp tests (SOFTWARE-5812)

matyasselmeci:
The dependency "osdf-client" is not available, except in 23-upcoming.

VMU run: https://osg-sw-submit.chtc.wisc.edu/tests/20240207-1110/packages.html. A spot check showed that the various "stashcache" tests are no longer getting skipped in the XCache test set.


mwestphall (approved):
LGTM

PR #269, opened 2024-02-08
Set STASH_USE_TOPOLOGY for stashcp tests for Pelican 7.5.0+ compat

matyasselmeci:
Pelican 7.5.0+ talks to the Director by default, not Topology, so the fake Topology server we're setting up in the tests gets bypassed. Setting STASH_USE_TOPOLOGY to non-blank makes it fall back to the previous behavior.

https://osg-sw-submit.chtc.wisc.edu/tests/20240208-1718/packages.html


mwestphall (approved):
LGTM

PR #270, opened 2024-04-11
Write grid-mapfiles for stashcache tests (SOFTWARE-5846)

matyasselmeci:
As of that ticket, we need grid-mapfiles in /run/stash-origin-auth and /run/stash-cache-auth. They are normally downloaded from Topology, but we don't hit topology in these tests.

Example run: https://osg-sw-submit.chtc.wisc.edu/tests/20240411-1529/packages.html


brianhlin (approved):
LGTM

PR #273, opened 2024-05-28
Fix scitokens debug level

matyasselmeci:
`scitokens.trace debug` only turned on 'debug' messages, not 'error', 'warning', and 'info'


brianhlin (approved):
LGTM

PR #274, opened 2024-10-24
Add 2X-main special casing to install

mwestphall:
- Code to convert `23` to `23-main` already existed in the "upgrade" logic, but was missing from the "downgrade" logic
- It might make more sense to remove the special casing and switch the parameters.d/*.yaml files to include `-main`, this might have wider code ramifications though
  - eg. `sources: [opensciencegrid:main; 23-main; osg, osg-upcoming > 24/osg-minefield]`

Successful test run using this patch: https://osg-sw-submit.chtc.wisc.edu/tests/20241024-1623/results.html


brianhlin (approved):
LGTM. I think the convenience in the param YAML / potential ramifications aren't worth removing this special casing

PR #277, opened 2025-05-06
Skip StashCache tests if pelican is installed

matyasselmeci:
"stash-cache" and "stash-origin" are deprecated (and have been removed in 24-main),
and the "pelican" command-line tool does not work with the osg-test
stashcache setup because it uses a director instead of topology.

At some point, separate tests for pelican should be added.


matyasselmeci:
VMU tests: https://osg-sw-submit.chtc.wisc.edu/tests/20250505-1618/packages.html
(as you can see, stash-cache and stash-origin can't be installed on OSG 24 anyway)

mwestphall (approved):
LGTM

PR #280, opened 2025-08-14
Accomodate change in daemon user for htvault-config

mwestphall:
Per Dave D, the daemon user changed from `vault` to `openbao` in 2.1.0 https://opensciencegrid.atlassian.net/browse/SOFTWARE-6198


brianhlin (approved):
LGTM. @mwestphall can you make sure that we communicate the change in daemon user in the release notes?

PR #288, opened 2025-10-02
Drop voms-server tests

matyasselmeci:
We don't install voms-server anymore


matyasselmeci:
https://osg-sw-submit.chtc.wisc.edu/tests/20251001-1924/packages.html

mwestphall (approved):

PR #290, opened 2025-10-10
Fix capturing stderr output

matyasselmeci:
Fixes a long-standing bug in `core.__run_command()` where stderr was always redirected to stdout, causing the `STDERR` sections in logs to always be empty.  Since some tests search through output looking for error messages, combine stdout and stderr explicitly in those tests.

https://osg-sw-submit.chtc.wisc.edu/tests/20251009-1714/results.html


mwestphall (approved):
LGTM

PR #292, opened 2026-07-13
Turn off the safety check that prevents TPCs from local or private addresses

matyasselmeci:
We're running in a VM with NAT networking, so this is our situation.

Tests: https://osgsw-ap2000.chtc.wisc.edu/tests/20260710-1551/packages.html


brianhlin (approved):
LGTM

PR #293, opened 2026-07-20
Stop using distutils

matyasselmeci:
It was removed in Python 3.12. osg-test itself doesn't even use it, it's just a leftover import. We do use it in the Makefile (though I don't know if we use the Makefile).


timtheisen (approved):
LGTM

